PR #42, opened 2014-07-16
bird: New version. Added lightweight client.

MartinStr:


sbyx:
Obsolete due to update to 1.4.5

PR #52, opened 2014-09-06
babeld: create the directory for the config file and restart babeld on ifup

caspermeijn:
babeld.init couldn't create the config file because of a nonexistent directory. This commit creates the directory before writing to the config file.


mwarning:
This PR doesn't seem to get anywhere. Any objections against closing it?

caspermeijn:
I agree. The first commit has been merged. 
For the second commit a proper fix is needed, the issue #131 describes the problem it is supposed to fix.

Issue #17, opened 2014-01-22
procd support for babel


kerneis:
Received on babel-users:

Over the past year openwrt has migrated over to an init replacement
called "procd"...

While the ideal situation is a daemon that never crashes, a better one
than presently is where babel is monitored and restarted properly. It
would be nice if babel could migrate over to USE_PROCD.

Regrettably the amount of rework required in babel.init is pretty
substantial, and it would be better if someone more intimate with
babel than I took it on.

a good example is in openwrt's repo for dropbear...

zorun:
Closed via #250

PR #55, opened 2014-09-11
babeld: procd support

imuli:
note: dependant on a change that hasn't been commited to procd yet,
adding support for null arguments.

fixes bug #17

### DIFF
--- a/babeld/files/babeld.init
+++ b/babeld/files/babeld.init
@@ -2,9 +2,9 @@
 
 . /lib/functions/network.sh
 
+USE_PROCD=1
 START=70
 
-pidfile='/var/run/babeld.pid'
 CONFIGFILE='/var/etc/babeld.conf'
 OTHERCONFIGFILE="/etc/babeld.conf"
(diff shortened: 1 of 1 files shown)


mwarning:
It would be nice if this could be merged. It seems to be ok from what I read here.

@kerneis bump :-)

jech:
I'll try to release 1.8 before the end of the summer, that will be a good time to merge this change.

zorun:
The change looks good to me, and it would be nice if this is merged before the LEDE release.

Can someone confirm that it indeed works in practice? @imuli ?  I can't test this easily, and I don't like merging without testing.

zorun:
The source of this pull request seems to have disappeared, so I opened a new pull request with Imuli's commit and some changes: #250

imuli:
@zorun I used that patch in production two years ago. It worked then, but I don't have the easy testbed or that job anymore!

zorun:
@imuli Ok, thanks :)

PR #66, opened 2014-11-27
nodogsplash: migrate to uci based configuration

lynxis:
Use /etc/config/nodogsplash for configuration.
Old setups will migrate using config option to use old
file base configuration.


lynxis:
I would like to support both.
any suggestions?

mwarning:
Let's support both then. :-)

mwarning:
Ok, merging now. Some minor things will be fixed afterwards.

PR #75, opened 2015-01-03
[olsrd] adapt init script for 0.6.8 (PingCmd option)

andrenarchy:
olsrd 0.6.8 accepts a PingCmd parameter as of
http://olsr.org/git/?p=olsrd.git;a=commitdiff;h=0a26c52af3941eb9060b193cae5552f8d13fd28e.
The PingCmd is a string of the form "ping -c 1 -q %s" where %s is
replaced with an IP. This commit allows for '%' characters in an olsrd
option.


zioproto:
thank you

PR #86, opened 2015-03-29
batman-adv with kernel 4.0

dangowrt:
batman-adv-2014.04 is broken since kernel commit af3cd13501 "lib/string.c: remove strnicmp()"


lindnermarek:
Sorry that I did not respond earlier. Somehow I don't get informed via email. Only @mwarning's hint brought me here.

I just updated the package to 2015.0 which should provide Linux 4.0 compatibility.

PR #88, opened 2015-04-16
nat46: Delete 464xlat device when pause is interrupted by SIGTERM

dedeckeh:
This avoids warning traces like "Can not add: device clat already exists!" when the clat interface is brought up again after previously being brought down


sbyx:
Thanks a lot. Next time please add a Signed-off-by.

PR #90, opened 2015-04-20
luci-app-cjdns: import package from SeattleMeshnet/meshbox

dangowrt:
Signed-off-by: Daniel Golle <daniel@makrotopia.org>


dangowrt:
As discussed with @lgierth the luci gui should also go to the routing feed just like the luci stuff for other routing protocols. @sbyx ?

dangowrt:
Thanks!

PR #91, opened 2015-04-20
nat46: Refuse link local address as implicit source in 464xlat

dedeckeh:
Signed-off-by: Hans Dedecker <dedeckeh@gmail.com>


sbyx:
thanks.

PR #92, opened 2015-04-21
cjdns: support libc other than uClibc

dangowrt:
Properly pass UCLIBC=1 only in case we are actually building with uClibc. Fixes SSP support at least on musl (and probably glibc).

Signed-off-by: Daniel Golle <daniel@makrotopia.org>

### DIFF
--- a/cjdns/Makefile
+++ b/cjdns/Makefile
@@ -36,8 +36,14 @@ endef
 define Build/Configure
 endef
 
+PKG_DO_VARS:=
+
 ifneq ($(CONFIG_KERNEL_SECCOMP_FILTER),y)
-PKG_DO_VARS:=Seccomp_NO=1
+PKG_DO_VARS+= Seccomp_NO=1
+endif
+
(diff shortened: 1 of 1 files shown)


lgierth:
@sbyx Could I get commit access? There are a couple more features in sight, regarding automated setup of IP tunnels, and interoperability with non-cjdns networks.

I would also like to integrate our buildbot with the repo's web hooks, so that we can continue to have automated smoketests for commits -- I'll get back to you about this within a few days once the buildbot is ready for the change.

I'm happy to open the buildbot up for other packages that can be tested in a Docker container, or ARM VM.

sbyx:
@jow- what do you think?

jow-:
ACK to both the PR and the commit access.

sbyx:
@lgierth please read and follow the usual boilerplate: https://github.com/openwrt/packages/blob/master/CONTRIBUTING.md

lgierth:
Read and acknowledged, thanks!

PR #94, opened 2015-04-30
Some uci defaults fixes

dangowrt:
rant-mode=on: Generally, Linux interface names should not leak into uci, uci interfaces should be used instead (i.e. lan instead of eth0 or br-lan)
cjdrouteconf should query this from netifd via ubus similar to
ubus call network.interface status "{ \"interface\" : \"lan\" }"

### DIFF
--- a/cjdns/files/cjdns.defaults
+++ b/cjdns/files/cjdns.defaults
@@ -18,7 +18,7 @@ EOF
 
   # make sure config is present (might fail for any reason)
   uci get cjdns.cjdns.ipv6 >/dev/null 2>&1
-  if [ $? -ne 0]; then
+  if [ $? -ne 0 ]; then
     exit 1
   fi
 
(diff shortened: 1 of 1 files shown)


dangowrt:
@lgierth

dangowrt:
... a netifd protocol handler is supposedly the most elegant way to solve this. So similar to batman-adv you should add an interface, e.g.:
uci set network.cjdnslan=interface
uci set network.cjdnslan.ifname='@lan'
uci set network.cjdnslan.proto='cjdns'

See https://github.com/openwrt-routing/packages/blob/master/batman-adv/files/lib/netifd/proto/batadv.sh

Not such a difficult thing to do that for cjdns as well (I hope and will find out)

dangowrt:
Ah, just found https://github.com/SeattleMeshnet/meshbox/tree/mgmt
What's the status on that?

PR #95, opened 2015-05-01
babeld: Added missing $IPKG_INSTROOT

kostko:
Without this automatic init script activation in imagebuilder fails.

### DIFF
--- a/babeld/files/babeld.init
+++ b/babeld/files/babeld.init
@@ -1,6 +1,6 @@
 #!/bin/sh /etc/rc.common
 
-. /lib/functions/network.sh
+. $IPKG_INSTROOT/lib/functions/network.sh
 
 START=70
 


kerneis:
Thanks for the pull request. I'm a bit concerned not to find it 
documented anywhere. Some scripts in base-files seem to use it as well, 
but not other script in openwrt-routing for instance. Do you have any 
reference? Would it make sense to add this to other packages in this 
repo as well?

kerneis:
Also remember to include a sign off line: https://dev.openwrt.org/wiki/SubmittingPatches (point 10). Even though it doesn't matter too much for such trivial patches.

kostko:
> Do you have any reference? Would it make sense to add this to other packages in this repo as well?

Yes, all init scripts must have this. The issue is that the init scripts are evaluated at firmware image generation time by the imagebuilder. And at that time, the root path `/` is the host's root, so accessing functions will fail. And this will cause init script activation to fail which will cause the generated firmware image to not run the specific services at startup and a manual `enable` call will be required.

Init script activation happens here:
* https://dev.openwrt.org/browser/trunk/target/imagebuilder/files/Makefile#L151

> Also remember to include a sign off line

I've fixed the commit now.

kerneis:
I'm almost convinced to merge this, but I'd just like to fully understand. In the link you gave, imagebuilder is executing *.postinst scripts, and there is no explicit postinst script for the babeld package. Are those automatically generated? Or is this a script that you wrote separately to configure the image for your purposes, and containing something equivalent to `/etc/initd./babeld enable`? In the latter case, wouldn't a firstboot script be a better way to reach the same result?

In any case, I think I'll accept the patch because it seems harmless, and potentially useful. I'm just trying to figure out why this pattern is not more widespread, and what use case it solves exactly. Thanks again for your time.

kostko:
Oh, it seems this has been changed in trunk. Check out the barrier breaker branch, it uses the `enable` call:

* https://dev.openwrt.org/browser/branches/barrier_breaker/target/imagebuilder/files/Makefile#L149

kerneis:
Merged, and should probably go to Chaos Calmer branch as well (but I cannot do this from the web interface).

PR #102, opened 2015-05-30
alfred: do not handle 'none' as an existing interface

mwarning:
For reference:

```
$alfred -h
  -b                                  specify the batman-adv interface
                                      configured on the system (default: bat0)
                                      use 'none' to disable the batman-adv
                                      based best server selection
```

### DIFF
--- a/alfred/files/alfred.init
+++ b/alfred/files/alfred.init
@@ -67,11 +67,11 @@ alfred_start()
 	append alfred_args "$args"
 	enable=1
 
-        config_get_bool start_vis "$section" start_vis 0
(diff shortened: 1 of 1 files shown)


mwarning:
Hm, alfred does not seem to start at this point when the interface is set to none.
/var/run/alfred.pid is present, but no alfred process. But "/etc/init.d/alfred start" is successfull.
Adding a sleep of 30 seconds in the init script makes it work.

mwarning:
ok, figured it out now. We have to wait for batmanif an interface. But ignore batmanif it is 'none'.

mwarning:
ping @simonwunderlich :-)

PR #105, opened 2015-06-23
Fix cjdns build on mac

cjdelisle:
Fixes 2 bugs in the build which only appear when building on a mac.
1. ar and ranlib are not specified so the platform ones are used by nacl build ... which happens to work on linux but not on mac
2. unless make-linux is explicitly specified as the build format, GYP will attempt to build an xcode project which breaks libuv build.

### DIFF
--- a/cjdns/Makefile
+++ b/cjdns/Makefile
@@ -20,9 +20,9 @@ PKG_NAME:=cjdns
 PKG_VERSION:=0.16
 PKG_RELEASE:=14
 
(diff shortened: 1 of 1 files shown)


lgierth:
:+1: Reworded and pushed as 98d7598890e754e61bdd75a5b5b55d07c518a2b6 and fdd1fbf6e1605e8d76603c79be7e1c3d9e9148a2

PR #126, opened 2015-08-09
alfred: not only wait for the interface, but also a link-local address

NeoRaider:
Changes between alfred 2015.0 and 2015.1 cause alfred to print an error
and exit when the interface to bind to is not completely set up (has no
non-tentative link-local address). As a workaround, wait for such an
address by reading /proc/net/if_inet6 before alfred is started.

In the long term, it would be nice to make alfred more robust against
missing interfaces (allowing alfred to start without the interface
existing, and keeping it working even when the interface is removed and
re-added while alfred is running).

Signed-off-by: Matthias Schiffer <mschiffer@universe-factory.net>

### DIFF
--- a/alfred/files/alfred.init
+++ b/alfred/files/alfred.init
@@ -37,6 +37,31 @@ wait_for_dir()
 	fi
 }
 
+wait_for_ll_address()
+{
+	local iface="$1"
+	local timeout=30
+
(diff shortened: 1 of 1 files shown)


simonwunderlich:
Bastian, did you do the changes you want? Should we close this pull request ...?

bittorf:
sorry, i just forgot - merged now

PR #132, opened 2015-08-16
babeld: fix parsing of list import_table

tcatm:
When parsing list options using config_load option_cb will be called
with additional options ending in _ITEM$x and _LENGTH. This ignores any
option containing _LENGTH or _ITEM as they can't possible be a valid
babeld option making a config like

    config general
        list import_table 23
        list import_table 42

work.

Signed-off-by: Nils Schneider <nils@nilsschneider.net>


zorun:
Thanks for the patch, I'll test it soon.

Of course, this issue would better be fixed upstream, see also:

https://lists.openwrt.org/pipermail/openwrt-devel/2014-August/027663.html
https://lists.openwrt.org/pipermail/openwrt-devel/2014-August/027673.html

PR #146, opened 2016-01-21
nodogsplash: fix init script

sayuan:


mwarning:
@sayuan sorry, did not notice this pull request until now.

PR #150, opened 2016-02-12
fix quoting of daemon name

booo:
The current quoting did not allow substitution and resulted in log messages with
the prefix "${DAEMON}[hotplug]".


lynxis:
@booo I would prefix the commit summary by the package or folder

PR #156, opened 2016-03-06
Replace == operator in test with = operator.

stspdotname:
Requested by bittorf in https://github.com/openwrt-routing/packages/pull/155


bittorf:
thank you!

PR #157, opened 2016-03-10
LuCI app for BMX7

rogerpueyo:
These commits add the luci-app-bmx7 package, the LuCI-based web interface for configuring the BMX7 routing daemon and visualizing the device status, mesh nodes, links, topology, etc. It is based on the luci-app-bmx6 package.


p4u:
Thank you for the work.

But most of the files are coming from luci-app-bmx6, it would be nice to preserve the historical commits instead of creating new content. If not, once we eliminate the luci-app-bmx6 package (when deprecated) we will loose this information. Something like this: https://stackoverflow.com/questions/16937359/git-copy-file-preserving-history#16937834

Do you think it is possible? 
Thanks.

rogerpueyo:
Hi,

I see your point but, honestly, I don't know how to do what you ask. I've read the post you linked but I can't see how it can help. Never mind about the pull request.

Cheers!

PR #167, opened 2016-03-25
batman-adv: Use the mac80211 compat layer

ecsv:
BATMAN_V (in the upcoming release) will use structures from the mac80211 package. These structures can be incompatible to the structures defined in the header of the kernel. This can lead to weird problems like stack overflows or memory corruptions.

To avoid this, the batman-adv has to be compiled against the mac80211 headers instead of the kernel headers. But this requires that the compatibility headers of compat-wireless are used. So the batman-adv compatibility layer has (mostly) be removed and instead the compat-wireless/mac80211 compat layer has to be used.


ecsv:
Close this pull request because it is now included in #177 (because otherwise the other pull request wouldn't work)

PR #192, opened 2016-06-05
batman-adv: remove DEPENDS:=kmod-crypto-core

lynxis:
kmod-crypto-core was removed in openwrt/trunk svn 46820 and all
kconfig's are now built-in.

Signed-off-by: Alexander Couzens <lynxis@fe80.eu>

### DIFF
--- a/batman-adv/Makefile
+++ b/batman-adv/Makefile
@@ -27,7 +27,7 @@ define KernelPackage/batman-adv
   URL:=https://www.open-mesh.org/
   MAINTAINER:=Marek Lindner <mareklindner@neomailbox.ch>
   SUBMENU:=Network Support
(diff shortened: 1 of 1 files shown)


lynxis:
@lindnermarek @bittorf @jow-

FreifunkUFO:
is it for that error message on my builds? :-)
Collected errors: * satisfy_dependencies_for: Cannot satisfy the following dependencies for batctl: * kmod-crypto-core *

lynxis:
correct. fixes that error.

jow-:
LGTM. The `kmod-cryto-core` looks like something that should get selected through `kmod-crypto-crc32c` anyway (in case it still exists in some environment).

PR #193, opened 2016-06-06
cjdns: use temporary directory within the build environment

jow-:
By default, cjdns' build system uses the system wide `/tmp` directory to store
intermediate build artifacts.

Unfortunately its build system fails to clean after itself, leaving thousands
of `jsmake-<hash>` directories behind, taking up precious inodes and massively
slowing down processes traversing the temporary directory:

    root@buildbot:~# find /tmp/ -maxdepth 1 -type d -name 'jsmake-*' | wc -l
    1581

Attempt to solve that problem by using the upstream-introduced
`CJDNS_BUILD_TMPDIR` environment variable in order to move the intermediate
artifacts from the system-wide `/tmp` to a temporary directory within the
package build dir which is properly deleted upon package rebuild cycles.

### DIFF
--- a/cjdns/Makefile
+++ b/cjdns/Makefile
@@ -18,7 +18,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=cjdns
 PKG_VERSION:=0.17
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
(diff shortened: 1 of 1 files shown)


jow-:
@lgierth

lgierth:
This LGTM, thank you :+1:

PR #195, opened 2016-06-08
nat46: Depend on IPV6 as IPv6 is kernel builtin module now

dedeckeh:
Signed-off-by: Hans Dedecker <dedeckeh@gmail.com>


sbyx:
Thanks.

PR #197, opened 2016-06-17
batman-adv: Fix speedy join in gateway client mode

ecsv:
Speedy join only works when the received packet is either broadcast or an 4addr unicast packet. Thus packets converted from broadcast to unicast via the gateway handling code have to be converted to 4addr packets to allow the receiving gateway server to add the sender address as temporary entry
to the translation table.

Not doing it will make the batman-adv gateway server drop the DHCP response in many situations because it doesn't yet have the TT entry for the destination of the DHCP response.


ecsv:
@lindnermarek @simonwunderlich It would be really nice to have this patch for gluon-2016.2

The patch was forwarded to b.a.t.m.a.n. a week ago as https://patchwork.open-mesh.org/patch/16376/

ecsv:
This patch is now based on #205

PR #202, opened 2016-06-27
miniupnpd-2.0

axet:
1.9 has bugs please update

https://github.com/openwrt/packages/issues/2889


fingon:
Thanks, updated also finally in the master.

PR #206, opened 2016-07-05
CC: batman-adv: Add reference counting + nullptr fixes

ecsv:
* Avoid nullptr dereference in bla after vlan_insert_tag
 * Avoid nullptr dereference in dat after vlan_insert_tag
 * Avoid tt_req_node list put for unhashed entry
 * Fix orig_node_vlan leak on orig_node_release
 * Fix non-atomic bla_claim::backbone_gw access
 * Fix reference leak in batadv_find_router
 * Free last_bonding_candidate on release of orig_node


ecsv:
@simonwunderlich @lindnermarek this is relevant for you

The patch "Fix non-atomic bla_claim::backbone_gw access" had to be backported (v2016.2 and v2016.1 had different return types for some functions). The rest are the vanilla upstream patches

PR #209, opened 2016-07-18
ndppd: fix source URL

ryzhovau:
Looks like http://www.priv.nu is offline for eternity.

[Tested on Entware-ng](https://github.com/Entware-ng/entware-routing/commit/16cd0418a3077a81018e335c2d73da2571bf1a44).


kerneis:
Thanks!

PR #212, opened 2016-07-29
zerotier: add initial package

mwarning:
A package for ZeroTierOne: https://github.com/zerotier/ZeroTierOne

### DIFF
--- a/zerotier/Config.in
+++ b/zerotier/Config.in
@@ -1,0 +1,24 @@
+menu "Configuration"
+	depends on PACKAGE_zerotier
+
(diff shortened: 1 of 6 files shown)


mwarning:
Any thoughts if this belongs here?

zorun:
This looks like a VPN software, and all VPN softwares so far are in the main package repository.  What makes you think it should belong here?

mwarning:
@zorun: I guess the main reason was because it relates to routing and I have access here.
But I will submit it to the package repo then.

PR #215, opened 2016-07-30
Adding Travis-CI integration

yanosz:
Hello folks,

After some problems with building packages for 15.01 I decided to set up an CI based on Travis for the 15.05.1-branch. The result is here: https://travis-ci.org/yanosz/packages-1

Please include this configuration and Travis in your system, too - it's annoying to build non-building packages

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,13 @@
+language: bash
+
+install:
+  - ./tests/test_ar71xx_15.05.sh
+
+script:
+  - make -C tests/OpenWrt-SDK-15.05.1-ar71xx-generic_gcc-4.8-linaro_uClibc-0.9.33.2.Linux-x86_64 world V=99 -j1
+
+branches:
+  only:
+    - for-15.05
+
+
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -0,0 +1,13 @@
+language: bash
```
ecsv: Cannot find "bash" in the language documentation of travis-ci. I would have expected c/cpp here

.travis.yml
```diff
@@ -0,0 +1,13 @@
...
+
+install:
```
ecsv: Here is something missing to make sure that wget is the newest version from precise:

    sudo: required
    install:
      - sudo apt-get update && sudo apt-get install wget

Or just switch to trusty via:

    dist: trusty
    sudo: false

lynxis:
@yanosz how about using the same script as https://github.com/openwrt/packages is using?

PolynomialDivision:
> @yanosz how about using the same script as https://github.com/openwrt/packages is using?

Yeah! Awesome idea.

yanosz:
> @yanosz how about using the same script as https://github.com/openwrt/packages is using?

Well, this PR is from 30 Jul 2016, packages starting using travis as of Jun 12, 2017.

I'm somewhat surprised, that this zombie-PR is alive, still. Nvm.

PR #241, opened 2016-11-30
batman-adv: add patches from 2016.4-maint 2016-12-02

ecsv:
* batman-adv: Install module in $INSTALL_MOD_DIR/updates/net/batman-adv
 * batman-adv: Revert "fix splat on disabling an interface"
 * batman-adv: Detect missing primaryif during tp_send as error
 * batman-adv: Check for alloc errors when preparing TT local data

As bonus fix the build with 3.18:

 * Add get_link_net compat code for Linux < 4.0.0


ecsv:
@lindnermarek I already had this one queued up for a Freifunk Firmware but it looks like other communities also started to use 2016.4 (and therefore are interested in these patches). They ended up reporting problems which are solved by the revert of "fix splat on disabling an interface".

So maybe it is better to simply add the maint patches to openwrt-routing

lindnermarek:
Merged! Thanks Sven!

PR #243, opened 2016-12-16
batman-adv 2016.5

ecsv:
@simonwunderlich @lindnermarek Please pull or let me know of any problem!

The batman-adv bug fixes mentioned in these changelogs were already part of pull request #241. The patches for them are now removed because they are already part of the official tarball.

batman-adv 2016.5
=================

 * support latest kernels (3.2 - 4.10)
 * change installation path back to $INSTALL_MOD_DIR/updates/net/batman-adv
 * add simple (re)broadcast avoidance
 * reduce multicast TT and TVLV update delays under B.A.T.M.A.N. V
 * support B.A.T.M.A.N. V throughput detection when using VLANs on top of WiFi
 * improve documentation of sysfs and debugfs files
 * bugs squashed:
   - detect missing primaryif during tp_send as error
   - fix re-adding of previously removed interfaces via rtnetlink
   - fix rare race conditions on interface removal
   - handle allocation error when generating TT responses

batctl 2016.5
=============

 * reimplement traceroute/ping commands in userspace without debugfs
 * switch interface manipulation from (legacy) sysfs to rtnetlink
 * coding style cleanups

alfred 2016.5
=============

 * support interface validity checks on systems without debugfs
 * remove debugfs check during batadv-vis startup
 * allow out-of-order txend packets during transmissions


FreifunkUFO:
@ecsv @simonwunderlich @lindnermarek so how is it?

PR #245, opened 2016-12-21
update babeld to version 1.8 which is compatible to 1.7.1 and adds a rw socket

christf:
this means that babeld (especially the interfaces it uses) may be reconfigured during runtime.

### DIFF
--- a/babeld/Makefile
+++ b/babeld/Makefile
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=babeld
-PKG_VERSION:=1.7.1
+PKG_VERSION:=1.8.0
 PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://www.irif.univ-paris-diderot.fr/~jch/software/files/
-PKG_MD5SUM:=2f71794d4e67f8a5352164ce33611549
+PKG_MD5SUM:=eb1c66c382e9181c418ebd84e52b5af2
 PKG_LICENSE:=MIT
 
 include $(INCLUDE_DIR)/package.mk


christf:
this may provide better alternatives to stopping/restarting babeld.
If this topic is of interest, I have created an init script for babeld that makes use of this socket functionality to add interfaces on the fly to a running babeld instance.
For whoemever is interested the file is available here: https://github.com/openwrt-routing/packages/pull/245
I am specifically not adding it to this PR in order to not overload it

christf:
@zorun @jech it seems you two ar teh go-to-people for the babeld package for openwrt - may this be merged?

jech:
> @zorun @jech it seems you two ar teh go-to-people for the babeld package for
> openwrt - may this be merged?

I haven't checked the MD5 -- but yes, 1.8 is compatible with 1.7, it
should get into OpenWRT.  @zorun, objection?

-- Juliusz

kerneis:
Looks good to me.



Gabriel

zorun:
@christf as mentionned in the Makefile, the main maintainer is @kerneis .

The change looks good to me.  The most important change in babeld 1.8.0 is this one:

```
  * Changed the configuration language to use an enumerated type instead
    of the "wired" boolean.  This is an incompatible change.
```

but babeld seems to accept the old syntax.  Also, option names are not hardcoded in the init script, so there's no change needed on our side.

zorun:
@christf Could you just change your commit message? (prefix with "babeld:", remove the part about the socket since you don't integrate it in this commit).  Also, can you add a Signed-off-by?

Thanks!

christf:
@zorun thank you for pointing this out. I only looked in the comments at the top. the commit message has been changed and signed.

zorun:
Merged, thanks!

christf:
thank you.

jech:
Thanks, Zorun.

PR #246, opened 2016-12-22
olsrd: disable compiler-warning "missing-include-dirs"

### LINKED ISSUE #244: olsrd build broken since commit 3b14a1c5b6aaf0422374a14998279d9047b0abf7
buildbot says
```
make[6]: Entering directory '/home/lede/buildbot/mips_24kc/build/sdk/build_dir/target-mips_24kc_musl-1.1.15/olsrd-0.9.5/lib/pud/nmealib'
[CC] src/sentence.c
cc1: error: /home/lede/buildbot/mips_24kc/build/sdk/staging_dir/target-mips_24kc_musl-1.1.15/include: No such file or directory [-Werror=missing-include-dirs]
cc1: all warnings being treated as errors
Makefile:57: recipe for target 'build/sentence.o' failed
make[6]: *** [build/sentence.o] Error 1
make[6]: Leaving directory '/home/lede/buildbot/mips_24kc/build/sdk/build_dir/target-mips_24kc_musl-1.1.15/olsrd-0.9.5/lib/pud/nmealib'
Makefile:162: recipe for target 'nmealib' failed
make[5]: *** [nmealib] Error 2
make[5]: Leaving directory '/home/lede/buildbot/mips_24kc/build/sdk/build_dir/target-mips_24kc_musl-1.1.15/olsrd-0.9.5/lib/pud'
Makefile:241: recipe for target 'libs' failed
make[4]: *** [libs] Error 2
```

SvenRoederer:
remove this warning as it fails to build for the nmealib with missing
"staging_dir/target-mipsel_24kc_musl-1.1.15/include" from buildenv

this fixes #244 on OpenWrt + LEDE master (checked via SDK-build)

### DIFF
--- a/olsrd/patches/missing-include-dirs-warning.patch
+++ b/olsrd/patches/missing-include-dirs-warning.patch
@@ -1,0 +1,12 @@
+diff -ur olsrd-0.9.5_orig/lib/pud/nmealib/Makefile.inc olsrd-0.9.5/lib/pud/nmealib/Makefile.inc
+--- olsrd-0.9.5_orig/lib/pud/nmealib/Makefile.inc	2016-12-22 12:23:42.175283967 +0100
++++ olsrd-0.9.5/lib/pud/nmealib/Makefile.inc	2016-12-22 12:24:24.591015066 +0100
+@@ -63,7 +63,7 @@
+                 -Wmissing-format-attribute -Wno-multichar -Wno-deprecated-declarations -Wendif-labels -Wwrite-strings \
+                 -Wbad-function-cast -Wpointer-arith -Wcast-qual -Wshadow -Wformat -Wsequence-point -Wcast-align \
+                 -Wnested-externs -Winline -Wdisabled-optimization -funit-at-a-time -fPIC -ggdb -Wformat=2 -Winit-self \
+-                -Wmissing-include-dirs -Wswitch-default -Wswitch-enum -Wconversion -Wdouble-promotion \
++                -Wswitch-default -Wswitch-enum -Wconversion -Wdouble-promotion \
+                 -Werror=format-security -Wformat-security -Wformat-y2k -Wredundant-decls -Wundef -Wunreachable-code \
+                 -Wunused-parameter
+ 


SvenRoederer:
An other option is too dig for the point the CXXFLAG is added or to just create the missing folder ...

bittorf:
@fhuberts - should we fix this upstream?

fhuberts:
upstream is always better.

but why is this needed? what warning do you get?

fhuberts:
ah, seems to me that the build environment is faulty here.

you add the dir but it doesn't exist.

jow-:
Unfortunately its the way LEDE and OpenWrt SDKs work, also existing released ones. They pass a number of include directories which might or might not exist.

In order to avoid the need for patching it downstream or to remove `-Wmissing-include-dirs` upstream can we get a variable which we can set in the LEDE/OpenWrt makefile? Something like `EXTRA_CFLAGS += -Wno-missing-include-dirs`?

fhuberts:
Well, that same flag is in release v0.9.5, so when you upgrade to that release you'll get the same warning from the rest of the olsrd build.

However, having just read the man page of gcc I'll just remove the warning, it doesn't seem that useful while causing trouble for you.

I'll fix it on master, since I'm planning for a new release after we have fixed another issue that we've encoutered.

bittorf:
@jow question: is it possible to just create the include-dirs when they are passed?

fhuberts:
fixed on olsrd master, see https://github.com/OLSR/olsrd/commit/7f809ea15f5caba43349d270c798e3f9ae9f2121

fhuberts:
BTW I'd appreciate it very much if you try to fix olsrd issues upstream. You might even consider cc-ing me on olsrd related issues...

jow-:
@bittorf - you can try to filter out `-I` arguments refering to not existing directories in the OpenWrt Makefile using something like this:

    INC_FLAGS := $(patsubst %,-I%,$(realpath $(patsubst -I%,%,$(filter -I%,$(TARGET_CFLAGS)))))
    TARGET_CFLAGS := $(filter-out -I%,$(TARGET_CFLAGS)) $(INC_FLAGS)

bittorf:
@jow I do not want to filter it out. the idea is that no interaction is needed for package-maintainers, so *if* the buildsystem "invents" direcories, it should at least create them.

jow-:
@bittorf - I was merely offering a pragmatic solution. Upstream code with string warning flags can always cause issues. Since neither patches to a not yet released olsrd nor patches to a not yet released OpenWrt or LEDE will address this I was pointing out a possibility to circumvent the problem locally in the olsrd Makefile.

PR #250, opened 2017-01-09
babeld: Add procd support and other fixes

zorun:
Based on the work of @imuli with some changes.

### DIFF
--- a/babeld/Makefile
+++ b/babeld/Makefile
@@ -12,7 +12,7 @@ PKG_VERSION:=1.8.0
 PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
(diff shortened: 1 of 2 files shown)


zorun:
@kerneis @jech any objection against merging this?

I refreshed the procd support from @imuli because of changes introduced since then, and babeld now logs to `logread` by default (instead of `/var/log/babeld.log`).  The rest consists of small fixes.

zorun:
By the way, I found a router running OpenWRT CC with babeld, so I tested these changes.  I should hope that it still works with newer versions of OpenWRT/LEDE.

kerneis:
Looks good to me, @zorun (although I have not tested it for real).
Thanks a lot for your help.


Feel free to update the maintainer to yourself if you wish. If you'd
rather not, I'm fine with keeping it as is ;-)


-- 

Gabriel

zorun:
@kerneis Ok, will do :)  I think I will add myself as maintainer then (but also keep you).

PR #253, opened 2017-01-15
Olsrd: remove 6and4 backward compatibility

SvenRoederer:
It's announced since several years (mid of 2014), that such setups (one olsrd running IPv4 and IPv6) will no be supported anymore. There seems to have been enough time to implement the required changes.

Comparing the generated configfile shows that they are equal. Only the IPv6-config has an "IpVersion 6" setting added now

### DIFF
--- a/olsrd/files/olsrd.init
+++ b/olsrd/files/olsrd.init
@@ -15,8 +15,6 @@ OLSRD_INTERFACE_DEFAULTS_SCHEMA='AutoDetectChanges:bool'
 T='	'
 N='
 '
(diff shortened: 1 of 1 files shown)


bittorf:
@thuehn any opinion about that? I remember there was an issue with running IPv4 + IPv6 with different deamons (pidfile?) - i'am OK with removing this old, deprecated feature...

thuehn:
Hi all,
I am absolutly fine with this change. 
Thx Sven

SvenRoederer:
I'm happy to get this whole script cleaned a bit

bittorf:
thanks sven, the script is really a mess 8)

PR #257, opened 2017-01-16
olsrd: nameservice: add postinst-hook to upgrade plugin

SvenRoederer:
olsrd-0.9.6 will come with an new versionnumber for the nameservice-plugin, as of new parameter "filewrite_interval".
This commit installs a postinst-hook for mod-nameservice to update the UCI-option library to the new value. When installing the same version nothing is done, on downgrades a warning is printed.

Without this the parameter will stay at "olsrd-nameservice.so.0.3". The init-script will check for this filename and will not include this plugin into the runtime-config.

### DIFF
--- a/olsrd/Makefile
+++ b/olsrd/Makefile
@@ -179,6 +179,54 @@ define Build/Compile
 	$(call Build/Compile/Default,libs)
 endef
 
+define Package/olsrd-mod-nameservice/postinst
+#!/bin/sh
+. "$${IPKG_INSTROOT}/lib/functions.sh"
+
+# gets lookup up on compiletime and path is stripped-off
+PLUGIN_CURRENT_VERSION=$(subst $(PKG_BUILD_DIR)/lib/nameservice/olsrd_nameservice.so.,,$(wildcard $(PKG_BUILD_DIR)/lib/nameservice/olsrd_nameservice.so.*))
+
+upgrade_plugin_options() {
+        local cfg=$$1
+        local version
+        local value
+
+        config_get value $$cfg library
+        if [ $${value:0:20} = olsrd_nameservice.so ]; then
+                #echo found nameservice
+                if [ $${value} = olsrd_nameservice.so.$${PLUGIN_CURRENT_VERSION} ]; then
+                        #echo section up-to-date
+                        return 0
+                fi
+                version=$${value/olsrd_nameservice.so./} # strip-off the library-name
+                major=$$(echo $$version | cut -d . -f 1)
+                minor=$$(echo $$version | cut -d . -f 2)
+                configured_ver=$$((major * 1000 + $$minor))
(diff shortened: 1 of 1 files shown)


bittorf:
i hate it when a package wants to 'uci commit'. please only show a warning. in our firmware we have something "automatic":

```
_olsr_plugin_filename_get()
{
        local name="$1"
        local file

        # sorts/prints highest number if multiple versions are installed (in one dir)
        for file in "/usr/local/lib/olsrd_$name.so."* "/usr/lib/olsrd_$name.so."* "/lib/olsrd_$name.so."*; do {
                [ -e "$file" ] && {
                        basename "$file"
                        return 0
                }
        } done

        echo "not_found:olsrd_$name.so.0.0"
        return 1
}

for PLUGIN in txtinfo nameservice jsoninfo arprefresh watchdog dyn_gw; do
 : warn if newest installed differs from uci-value
done
```

SvenRoederer:
Yeah, it doesn'T feel really cool.
But the idea was a remote site only connectign via OLSR to the net. When upgrading and not changing the library-filename, the daemon will not start, so a site-visit needs to be done.

bittorf:
why will the daemon not start? where exactly is the error? the old library is still in place, is'nt it?

SvenRoederer:
indeed, the daemon should start up with it's basic features.
But the plugin will not load; I'll check again, what caused to daemon to fail in my setup.

The old lib gets removed, when opkg installs the updated plugin-package.

bittorf:
so we should better make the daemon startup but show a warning (`logger -s ...`) in log. please check what is the issue...(i'am a little bit blocked till end of the week)

SvenRoederer:
Did a bit testing and found that the current way of upgrade / startup only disables the plugin, that have the version number changed. Such plugin will not be configured and a message will be printed to console (only?). So it's not causing the  daemon to fail startup.
The core-functionality (setup the routing) will not be affected and the networking should be fine.
The user will find some additional functionality not working and we should ensure, that he finds some helpful logging about the missing plugin. And he should check the olsrd-changelog what changed and correct the plugin-name option and possibly add / remove changed options.

@bittorf thanks for asking "why will the daemon not start?" So it's that the daemon will fail, but only some plugins will not be activated.
If you agree, that full automatic updates are not scope of this package, you can close this PR

bittorf:
we can introduce a new uci-var "automatically use newest plugins" and on each start iterate over "_olsr_plugin_filename_get()" and set the correct plugin-names?

SvenRoederer:
newer plugins will usually bring new parameters. This will soon lead to the fact that such autoupgrade will "cry" for some "setup new defaults"-requirement.
Not sure if it makes sense set the defaults from this feed. different use-cases might require different default settings, which will break one or the other side.
For new setups / installation defaults are fine, but upgrading ... I tend to "have the operator setup his settings for new paramerters"

bittorf:
I agree, we will keep things as they are - KISS 8-)

PR #258, opened 2017-01-18
ndppd: fix compilation with modern gcc

jow-:
Import upstream commit "f19fa4b Fix a cast so it can be compiled with gcc-6"
in order to fix building ndppd with recent gcc versions.

Fixes the following error spotted by the build bots:

    src/iface.cc: In static member function 'static ndppd::ptr<ndppd::iface> ndppd::iface::open_pfd(const string&)':
    src/iface.cc:153:5: error: narrowing conversion of '-1' from 'int' to '__u32 {aka unsigned int}' inside { } [-Wnarrowing]


jow-:
@kerneis - please consider merging this.

Upstream commit: https://github.com/DanielAdolfsson/ndppd/commit/f19fa4be6ac87032c01f2db57ccb484f83c62604

PR #259, opened 2017-01-18
batmand: switch to tarball download, fix build with musl libc

jow-:
Since the upstream SVN repository is not available anymore, update the
Makefile to fetch tarballs cached by the LEDE or OpenWrt projects instead.

Also add a missing -D_GNU_SOURCE to the target flags to fix compilation
with musl libc.


jow-:
@lindnermarek - please consider merging this PR.

PR #134, opened 2015-08-19
batmand: use GNU_SOURCE for musl

plntyk:
fixes GCC5

enable extensions to fix errors like:
posix/tunnel.c:95:82: error: 'struct udphdr' has no member named 'dest'

Signed-off-by: Dirk Neukirchen dirkneukirchen@web.de

PR #261, opened 2017-01-23
batman: Switch maintainers and get batmand/vis packages from git

ecsv:
The MAINTAINER fields which were set to Marek Lindner was changed to:

 * batctl/batman-adv: Simon Wunderlich <sw@simonwunderlich.de>
   - Marek asked Simon to handle the pull requests/package updates directly
 * batmand/vis: Corinna "Elektra" Aichele <onelektra@gmx.net>
   - Marek clarified that Elektra is the maintainer: https://github.com/openwrt-routing/packages/pull/134#issuecomment-236346560

batmand/vis were splitted into two different source packages. This is necessary to get the sources directly from git and not from the LEDE mirror. This is (more or less) a follow up to #259. It doesn't upgrade these two packages (which are still on the same version from mid-2009).


ecsv:
@simonwunderlich Is it ok to do these things together or should I split the pull request?

PR #263, opened 2017-01-26
Add Prince to openwrt-routing

gabri94:
Hi, i would like to add my [implementation](https://github.com/gabri94/poprouting) for Pop-Routing to openwrt.
It is a demon used to optimize the timer's value of OONF, so i think it belongs to the routing feed.
I developed it as GSoC 2016 project with Freifunk and Ninux. 

Unfortunately due to a C++ Library the v0.1 version wasn't installable on embedded devices. (2MB of size)
Recently we've re-implemented this library in pure C and the new version (v0.2) has a tinier packet (20KB).

Thanks, Gabriel

### DIFF
--- a/poprouting/Makefile
+++ b/poprouting/Makefile
@@ -28,8 +28,9 @@ define Package/prince/description
 Prince is an open source implementation of the PopRouting Algorithm.
 It has been developed as a Google Summer of Code Project in collaboration with Freifunk and the University of Trento.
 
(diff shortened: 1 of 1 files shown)


bittorf:
thank you! this is untested on my side - but is does not break anything, so i merge.

gabri94:
May you merge it also in the lede and 15.05 feed?  Thanks

bittorf:
you have to make a PR for this...

gabri94:
Hi, I updated my package. Should I create a new PR?

PR #266, opened 2017-01-27
batman-adv: use kernel-specific PKG_BUILD_DIR

NeoRaider:
@ecsv @lindnermarek

Please also apply to lede-17.01.

The batman-adv Makefile set a PKG_BUILD_DIR outside of KERNEL_BUILD_DIR;
this could lead to batman-adv not being rebuilt when switching targets and
thus kernel configurations. This resulted in ABI mismatches and
occasionally broke the build with messages like the following:

    Package kmod-batman-adv is missing dependencies for the following libraries:
    crc16.ko

Instead of setting a better PKG_BUILD_DIR ourselves, we can just remove
the explicit PKG_BUILD_DIR definition and include kernel.mk before
package.mk to get the default definition used by other kernel module
packages.

Reported-by: David Lutz <kpanic@hirnduenger.de>
Signed-off-by: Matthias Schiffer <mschiffer@universe-factory.net>

### DIFF
--- a/batman-adv/Makefile
+++ b/batman-adv/Makefile
@@ -11,17 +11,15 @@ include $(TOPDIR)/rules.mk
 PKG_NAME:=batman-adv
 
 PKG_VERSION:=2016.5
-PKG_RELEASE:=0
+PKG_RELEASE:=1
 PKG_MD5SUM:=6717a933a08dd2a01b00df30cb9f16a8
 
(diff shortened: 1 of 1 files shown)


ecsv:
@NeoRaider Marek isn't accepting pull requests anymore and I don't have access right.

So lets trigger @simonwunderlich :)

PR #267, opened 2017-01-29
batman-adv: compile batman v by default

lindnermarek:
@simonwunderlich @ecsv

Compile BATMAN V per default to make the routing algorithm available in the snapshot builds and release packages.

Signed-off-by: Marek Lindner <mareklindner@neomailbox.ch>


ecsv:
@lindnermarek I don't have write access to the repo

lindnermarek:
@ecsv Included you in case you had a comment and/or any objection.

PR #275, opened 2017-02-12
miniupnpd: add IGDv2 support as build option

kdarbyshirebryant:
Make IGDv2 support a build time option (default enabled)
Some devices (xbox one, windows home server) are not happy talking to an
IGDv2 enabled upnp server.

Remove 'portinuse' & 'igdv2' enabling patches backported from freebsd as
these can be configured via build time options in package makefile.

Signed-off-by: Kevin Darbyshire-Bryant <kevin@darbyshire-bryant.me.uk>


kdarbyshirebryant:
@fingon ping

PR #280, opened 2017-02-26
lede-17.01: batman-adv: add patches from 2016.5-maint 2017-02-21

ecsv:
* batman-adv: Fix double free during fragment merge error
 * batman-adv: Fix transmission of final, 16th fragment


ecsv:
@simonwunderlich

PR #285, opened 2017-03-02
For 15.05

gabri94:
#283


poprouting/Makefile
```diff
@@ -1,16 +1,16 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=prince
-PKG_VERSION:=v0.2
-PKG_RELEASE:=2
+PKG_VERSION:=v0.3.1
+PKG_RELEASE:=4
```
zioproto: You are updating also PKG_VERSION, so this PKG_RELEASEs should be 1.
PKG_VERSION follows the upstream, and PKG_RELEASE is used to track the packaging versions for the same upstream code release. Please fix this

gabri94:
fixed

PR #288, opened 2017-03-07
prince: renamed folder

gabri94:
Renamed the package's folder to match the package name.

Signed-off-by: Gabriele Gemmi <gabriel@autistici.org>

PR #289, opened 2017-03-07
prince: renamed folder

gabri94:
#288 

Signed-off-by: Gabriele Gemmi <gabriel@autistici.org>


mwarning:
See https://github.com/openwrt-routing/packages/pull/290

PR #308, opened 2017-07-03
OONF 0.14.1-1

HRogge:


mwarning:
@HRogge what is it with all these merge requests of the same name? Are they duplicates?

HRogge:
If I understand your question correctly, they are for different releases of openwrt, which all have (according to my knowledge) different branches in GIT.

mwarning:
The target branches are different. I did not notice that. :-)

HRogge:
Ahh, THAT explains the comment...

I tend to make one commit and then cherry-pick it for the other branches.

PR #310, opened 2017-07-03
miniupnpd: disable IGDv2 by default

jow-:
The upstream project also reverted IGDv2-by-default due to widespread
compatibility problems.

So far all Microsoft operating systems up to Windows 10, Xbox 360, Xbox One
Playstation 3 and Playstation 4 consoles seem to be incompatible to the
new 2.0 standard.

### DIFF
--- a/miniupnpd/Makefile
+++ b/miniupnpd/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=miniupnpd
 PKG_VERSION:=2.0.20170421
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE_URL:=http://miniupnp.free.fr/files
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
@@ -32,7 +32,7 @@ endef
 define Package/miniupnpd/config
 config MINIUPNPD_IGDv2
 	bool
-	default y
+	default n
 	prompt "Enable IGDv2"
 endef
 


jow-:
ping @fingon

fingon:
Looks good; I personally only use the PCP part so I do not have opinions one way or the other.

jow-:
thanks! :)

hnyman:
I think that this should also be backported to 17.01 that has IGDv2 always on (without config possibility)

17.01 misses the introduction of the IGDv2 config item in #275 , the later version bump and this change to the default IGDv2 value

fingon:
@hnyman Agreed. I will do little bit of backporting later today once I get home.

diizzyy:
This should rather be split into two packages.

fingon:
I got demotivated on 17.01 branch effort as I do not have test setup for it anymore (except production node). Pull requests welcome ;) 

@diizzyy Why? miniupnpd (by design) is pretty monolithic, and if IGDv2 stuff can be chosen by option is better in my opinion than two packages with (minor) build option differences.

diizzyy:
Packages as in binaries, not as in Makefile (if possible).

kdarbyshirebryant:
Something like https://github.com/kdarbyshirebryant/routing/commit/d06e0bb83fd01c6cb9a241af7a46eea15c6a5a9b would probably do the trick.  Lightly tested.  Needs checking by person with clue.

treysis:
I think this should be changed again to include IGDv2. Issue was the major version of UPnP being reported as "2", not the IGD version. IGDv2 is needed for IPv6 firewall pinholing. This is the way it has been changed by the upstream package: change major version to 1, minor to 1, IGDv2 enabled. Actually the miniupnpd from 18.06 (version 2.1) works fine on LEDE as well as far as I can tell.

PR #313, opened 2017-07-07
miniupnpd: igdv1 & igdv2 compatibility as separate packages

ldir-EDB0:
Build miniupnpd as two package variants, default is IGDv1
compatibility, igdv2 variant enables IGDv2 compatibility.

IGDv2 has backwards compatibility issues, notably with Microsoft
products and whilst miniupnpd can be built with IGDv2 disabled, until
recently the default had it enabled.  Two package variants allow the end
user to choose without getting into building own packages.

Signed-off-by: Kevin Darbyshire-Bryant <kevin@darbyshire-bryant.me.uk>


ldir-EDB0:
Will someone *PLEASE* look at this PR

ldir-EDB0:
good idea/bad idea?

zorun:
This looks like a good idea in principle, but I don't know much about build variants.

@fingon any opinion?

PR #315, opened 2017-07-13
nodogsplash2: initial package

mwarning:
This package is a transitional package for nodogsplash, so that the old version is still available for those who miss features that have not yet been implemented.

Signed-off-by: Moritz Warning <moritzwarning@web.de>

### DIFF
--- a/nodogsplash2/Makefile
+++ b/nodogsplash2/Makefile
@@ -1,0 +1,61 @@
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=nodogsplash2
+PKG_FIXUP:=autoreconf
+PKG_VERSION:=2.0.0
+PKG_RELEASE:=1
+
+PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)-$(PKG_VERSION)/
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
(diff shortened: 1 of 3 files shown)


mwarning:
@lynxis please take a look

nodogsplash2/Makefile
```diff
@@ -0,0 +1,59 @@
...
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=nodogsplash2
+PKG_FIXUP:=autoreconf
+PKG_VERSION:=2.0.0
+PKG_RELEASE:=1
+
+PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)-$(PKG_VERSION)/
```
lynxis: PKG_MAINTAINER tag is missing
PKG_MIRROR_HASH is missing

nodogsplash2/files/nodogsplash.init
```diff
@@ -0,0 +1,393 @@
...
+  }
+
+  echo " Testing required modules"
+
+  do_module_tests "ipt_mac"
+  do_module_tests "ipt_mark"
+
+  # test for imq modules, only if TrafficControl is enabled in conf
+  if ( grep -q -E '^[[:space:]]*TrafficControl[[:space:]]+(yes|true|1)' "$CONFIGFILE" ) ; then
+    do_module_tests "imq" "numdevs=2"
```
lynxis: I think we should remove the module tests since they only checks for the module, but not for built-ins.
I'm unsure if we really should test `iptables` by modifing the firewall. But if we do it, we should at least use our own Chain instead of INPUT / OUTPUT ...

mwarning:
@lynxis done

PR #317, opened 2017-08-04
nodogsplash2: fix install folder to match the orignal nodogsplash pac…

mwarning:
…kage


mwarning:
@bluewavenet I will merge as soon I was able to verify that it is working. Some time today..

Issue #299, opened 2017-06-04
quagga-ospf6d: doesn't get routes from other routers


suedwestlicht:
Maintainer: acinonyx

Environment:
LEDE Reboot 17.01.1 r3316-7eb58cf109 / LuCI lede-17.01 branch (git-17.100.70571-29fabe2)
TP-Link TL-WR1043N/ND v4

Description:
ospfd of quagga works ok.

ospf6d of quagga recognizes its neighbours but doesn't get routes.
Package: quagga-ospf6d
Version: 1.1.0-1

It seems we have a bad quagga version.
Please see https://lists.gt.net/quagga/users/32153

Would it be possible to package an older version of quagga?

lucize:
can you try this #319  ?

suedwestlicht:
How can I install the new version?

lucize:
`git pull origin pull/319/head` in git root of routing master tree
and after that make a symlink of frr into package/feeds/routing/
`make menuconfig` and you should find it in the routing submenu

lucize:
you need to setup a SDK  [hint](https://lede-project.org/docs/guide-developer/the-source-code)
clone the git, resolve dependencies based on your distro then

`./scripts/feeds update -a`
`./scripts/feeds install -a`
clone the PR, make the symlink
`make menuconfig `
select frr packages

lucize:
https://lede-project.org/docs/guide-developer/start

suedwestlicht:
git clone https://git.lede-project.org/source.git
cd source
./scripts/feeds update -a
./scripts/feeds install -a

*** all went ok ***

Now I am stuck. In which dir shall I clone what?

lucize:
cd feeds/routing
git pull origin pull/319/head
ln -s ../../../feeds/routing/frr/ ../../package/feeds/routing/
cd ../..
make menuconfig

suedwestlicht:
ok. symlink looks good (resolves)
make menuconfig (shows the correct architecture Atheros AR7xxx/AR9xxx)

now I am too stupid to finde the frr packages. How can I activate them? I guess I have to build quagga's ospf6d only?

do I have to check
[ ] Build the LEDE SDK
[ ] Package the LEDE-based Toolchain
?

lucize:
you're not stupid :) !
 on subtarget find your router model and version, confirm on the router box that you have the V4 and select the V4 from subtarget
[...]

PR #319, opened 2017-08-13
frr: add FRRouting, a Quagga fork

lucize:
this version is the initial release for LEDE/OpenWRT

Mantainer: ME
Compile/Run Tested: x86_64, sunxi, mips

Signed-off-by: Lucian Cristian <lucian.cristian@gmail.com>


lucize:
@jow- frr reached 3.0 stable, can we have this in ?
I'll try to finish backporting eigrp, 3.0 already has nhrp so maybe soon we can replace some CISCO routers by LEDE
Also Quagga is a bit outdated

lucize:
@diizzyy @acinonyx someone an input ?

diizzyy:
I'm not a user but how does this compare to quagga (which still seems active)?
FFRouting seems also like a quite new project as far as I can tell ( https://www.linux.com/news/2017/4/welcoming-frrouting-linux-foundation ) and I'm not sure what more it offers running on OpenWRT/LEDE compared to bird/quagga except being yet another alternative?

lucize:
Quagga is quite stalled also [see statement](https://www.theregister.co.uk/2017/04/04/quagga_open_source_routing_resuscitated_as_free_range_routing/)  and the reccent issue #299 solved by this version
To be honest 3.1 will be a major update to quagga, there is much development happening [see](https://github.com/FRRouting/frr/commits/master).
For now 3.1 has some problems with cross compile regarding python-devel needed on the host but the problem is planned to be addressed

acinonyx:
Please, also read https://lists.gt.net/quagga/users/32627 . @pjakma is expressing concerns on how FRR distributes Quagga source code.

lucize:
but I don't see any detail about that, as they further discussed on the mailinglist and I don't think that we should concern about that, if that will happen then FRR should take care of that
that means that I should not use the single EIGRP alternative out there because someone in the mailing list has some concerns without showing actual data ?

pjakma:
On Sun, 22 Oct 2017, Lucian CRISTIAN wrote:

> but I don't see any detail about that, as they further discussed on 
> the mailinglist and I don't think that we should concern about that, 
> if that will happen then FRR should take care of that that means that

Quagga twice received opinion from lawyers.

The first relatively informal. Some, who then went on to push for a fork 
once they did not get their way, would not accept that opinion. They 
formulated various techy-pseudo-legal arguments, e.g. that copyright 
works differently for source versus binaries.

A formal legal opinion was then obtained for the Quagga project, with 
the query to the solicitor drawn up based on arguments from both sides. 
(note: I know FRR people tried - and failed - to obtain legal opinion 
favourable to their position before then).

The legal state is, as far as I know, clear: source code of derived 
works of Quagga must be distributed under the terms of the Quagga 
licence, i.e. the GPLv2+.

Those leading the FRR project have knowingly and deliberately chosen not 
to do that. As a result, Section 4 of the GPL on the code from Quagga 
applies to the FRR project, to Cumulus Networks and their agents, and 
other associated parties: their GPLv2 licence to the code base has 
terminated.

They have made arguments about the licence on the Quagga code that would 
permit distribute proprietary derived works from the Quagga codebase.

I don't intend to allow those arguments to stand.

regards,
-- 
Paul Jakma | paul@jakma.org | @pjakma | Key ID: 0xD86BF79464A2FF6A
Fortune:
You will have good luck and overcome many hardships.

pjakma:
Note: That the legal issues with their desired form of distribution
would end up causing difficulties for downstream distributors was 
pointed out to parties on the FRR side, on several occasions, over 
years.

regards,
-- 
Paul Jakma | paul@jakma.org | @pjakma | Key ID: 0xD86BF79464A2FF6A
Fortune:
No violence, gentlemen -- no violence, I beg of you!  Consider the furniture!
 		-- Sherlock Holmes

pjakma:
Given someone decided to link to The Register article that is very one-sided (and contains highly misleading statements):

https://lwn.net/Articles/718859/
https://lists.quagga.net/pipermail/quagga-dev/2017-April/033225.html

Missing from the above, that I could not state until recently, due to implicit threats by my former employer:

https://lists.quagga.net/pipermail/quagga-dev/2017-October/033285.html

lucize:
I don't want to state who's whom, weren't on the loop with things, this is just a Makefile that is using the source code available on github having more options from other open source projects

side notes: this is thing is not good for anyone, and if legal actions needs to be made, as an end user, should we just wait and see **IF** someday this will be addressed ?

lucize:
Updated to stable 5.0.1 there are some small changes to be done and it will be ready

Hopefully now has a clear advantage over the existing alternatives

cfra:
I have tested this package, and all in all, it seems to work well.

There are some smaller nits which I found:

a) `more` is absent, a different pager needs to be used. Trying out the different options, I found that `cat` was indeed the most usable option, since busybox's `less` apparently always start out with clearing the screen, which makes interactive use very annoying

The pager can be set with configure by now, I have added `--with-vtysh-pager=cat` to the configure statement and it works fine now. If one really wants to go with `less` at least `less -F -~` should be chosen so that it doesn't wait if the output fits on a single screen.

b) `ldpd` is not in the daemon list in `/usr/sbin/frr.init`, so it is not getting started automatically

c) `pbrd` is missing from the package

pjakma:
Hi,

The FRR project vehemently denies the applicability of the GPL licence 
of code obtained via Quagga (and GNU Zebra before that) to other code 
they distribute, where that other code has been made to be explicitly be 
heavily dependent on said GPL code for its function and comprehension, 
e.g. ldpd and babeld. And they encourage others to take the same 
position too.

They refuse to acknowledge the applicability of the GPL to that other 
code by deliberately refusing to put GPL notices on the code, as is 
customary, when that code carries notices of other licenses (which do 
not fulfill the conditions required by the GPL).

This is a very deliberate position of the FRR project, which has 
been expressed consistently over a number of years - and which they 
acted on from day-0 of the fork (indeed, this was the driving force for 
the fork, as far as I am concerned). E.g.:

   https://github.com/FRRouting/frr/issues/1923

The consequence is this, as far as I am concerned:

- The termination provisions of the GPL licence became applicable to the
   FRR project, including (but not limited to) NetDEF,
   OpenSourceRouting.org, Cumulus Networks, 6WIND, Big Switch Networks.

- No GPL licence is available for the affected code.

- Distribution or use by of the affected code is also unlicensed.

That means downstreams of FRR have no licence to use or distribute (and 
saying "well, whatever about FRR, /I/ choose to recognise the GPL as 
applying to that code, and I'll claim to distribute it under the GPL" - 
won't fix things, particularly not if the code is distributed exactly as 
FRR do).

This situation can not be fixed. Certainly not so long as the principals 
of the FRR project refuse to acknowledge that any code that explicitly 
is built on and depends on GPL code must itself be distributed under 
terms that *fully* implement the conditions required by the GPL licence.

If the OpenWRT project wants to violate the copyright of code I made 
available under the GPL via the Quagga project, then the OpenWRT project 
will be at risk of having its licence to my code terminated.

I reserve the right to recover damages and/or compensation for any use 
of software of mine, that is not otherwise stricly in accordance with 
the licence conditions I made it available under.

That includes from distributors such as the OpenWRT project, as well as 
any users of OpenWRT.

regards,

Paul

On Thu, 23 Aug 2018, Lucian CRISTIAN wrote:

> Updated to stable 5.0.1 there are some small changes to be done and it will be ready
>
> Hopefully now has a clear advantage over the existing alternatives
>
>

-- 
Paul Jakma | paul@jakma.org | @pjakma | Key ID: 0xD86BF79464A2FF6A
Fortune:
Dreams are free, but there's a small charge for alterations.

lucize:
@cfra updated, maybe you can see why the syslog repeats

`restart /usr/sbin/frr.init watchrestart process <x> exited with non-zero status 2`

lucize:
@cfra  cherry-picked

pjakma:
On Mon, 21 Jan 2019, Paweł Dembicki wrote:

> What is current legal status of frr? I saw some licenses changes 
> commits durring last time.
>
> Is still license problem?

If someone decides to distribute GPL code as BSD or MIT/X11 licensed 
code, against the advice of copyright holders in the GPL portions (who 
also gave constructive advice on alternate ways to do this, that would 
/not/ cause a legal mess - advice which was ignored), because of their 
own commercial imperatives, then the GPL licence /terminates/ for those 
people/entities.

Those who have lost their licence can not regain a licence by unilateral 
changes on their side. Those they distribute the software to also lack a 
licence to use or distribute the software.

----

I made code available to people, building on Kunihiro Ishiguro's (et al) 
GNU Zebra, under the GPL. I didn't ask for anything except that 
contributions be reviewed, and that the licence on my code (and /all/ 
others) be respected.

Some people were displeased by that. They smeared me. They even went to 
my employer, using their commercial connections with said employer. I 
was *barred* from working on a project I loved, even in _my own time in 
my own home with my own equipment_ - under threat of having my 
employment terminated.

There is no way, after that was done, that I am just going to forgive 
and forget, so long as the damage done is left unrepaired.

Note well: I reserve the right to recover compensation and damages from 
any parties involved in this infringement, either through distribution 
or use. Including from OpenWRT, should OpenWRT choose to distribute 
the infringing software.

regards,
-- 
Paul Jakma | paul@jakma.org | @pjakma | Key ID: 0xD86BF79464A2FF6A
Fortune:
You ain't learning nothing when you're talking.

CHKDSK88:
@pjakma 

Sorry, but I don't understand. FRR is distributed under GPL.  I'm confused.

c-x-berger:
> Those who have lost their licence can not regain a licence by unilateral
changes on their side.

Interesting.

The [closest thing Quagga has to a working homepage right now](http://savannah.nongnu.org/projects/quagga) specifies it is available under "GNU General Public License v2 or later".

This might be convenient for the FFRouting folks, since GPLv3 _explicitly_ re-grants license rights in Section 8, paragraph 2:

> However, if you cease all violation of this License, then your license from a particular copyright holder is reinstated (a) provisionally, unless and until the copyright holder explicitly and finally terminates your license, and (b) permanently, if the copyright holder fails to notify you of the violation by some reasonable means prior to 60 days after the cessation.

I say "might" because FFRouting is GPLv2, not v3, and I have no idea if distributing a derivative of v3 software as v2 is legal.

PR #326, opened 2017-10-13
[olsr] Bump olsr version to 0.9.6.2

thuehn:
This patch bumps olsr version from 0.9.6.1 to 0.9.6.2
One major bugfix comes with the newer version, which is quite relevant
for our Freifunk mesh network: the feature LinkQualityMult was not
working in 0.9.6.1 and got fixed in the new version.

Tested on 92 productive mesh nodes running current LEDE trunk .

Signed-off-by: Thomas Huehn <thomas@net.t-labs.tu-berlin.de>


thuehn:
comments ?

bittorf:
sorry for the delay!

PR #327, opened 2017-11-10
CC: ndppd: fix source URL

wigyori:
Maintainer: @kerneis 
Fix source URL for the 15.05 branch.


mwarning:
Thanks for fixing such an old source.

PR #332, opened 2017-12-05
batman-adv 2017.4

ecsv:
@simonwunderlich 

batman-adv 2017.4
=================

* support latest kernels (3.2 - 4.15)
* coding style cleanups and refactoring
* documentation cleanup
* bugs squashed:

  - avoid spurious warnings from bat_v neigh_cmp implementation
  - fix check of gateway availability in B.A.T.M.A.N. V
  - fix locking for bidirectional TQ check counters
  - remove leak of stack bits in fragmentation header priority

batctl 2017.4
=============

* synchronization of batman-adv netlink header
* coding style cleanups and refactoring
* documentation cleanup
* bugs squashed:

  - improve error handling for libnl related errors
  - add checks for various allocation errors

alfred 2017.4
=============

* synchronization of batman-adv netlink header
* coding style cleanups and refactoring
* documentation cleanup
* bugs squashed:

  - only query debugfs when netlink failed
  - handle allocation errors in hashtable iterator

PR #337, opened 2017-12-26
batman-adv: Fix missing function name for timer_setup in compat-hacks

ecsv:
@simonwunderlich


ecsv:
@mwarning might also be interested because he merged the pull request #332

Issue #334, opened 2017-12-14
batman-adv 2017-4: fails to compile with LEDE17.01


SvenRoederer:
For freifunk-berlin I recently build the master-feed against LEDE-17.01 (2b66449), which fails with
```
  CC [M]  /mnt/hosts/build/src/openwrt/freifunk/lede/lede/build_dir/target-mipsel_24kc_musl-1.1.16/linux-ramips_mt7620/batman-adv-2017.4/net/batman-adv/../../compat-sources/net/core/skbuff.o
In file included from <command-line>:0:0:
/mnt/hosts/build/src/openwrt/freifunk/lede/lede/build_dir/target-mipsel_24kc_musl-1.1.16/linux-ramips_mt7620/batman-adv-2017.4/compat-hacks.h:324:22: error: expected identifier or '(' before 'struct'
 static inline void  (struct timer_list *timer,
                      ^
```

Batman-adv 2017-4 was merged in https://github.com/openwrt-routing/packages/pull/332

p4u:
Same problem here.

nzamps:
Try changing the define in combat-hacks.h to :

```
static inline void      timer_setup(struct timer_list *timer,
                               void (*callback)(struct timer_list *),
                               unsigned int flags)
```

SvenRoederer:
@nzamps this patch let it build with LEDE-17.01 / https://github.com/lede-project/source/commit/ceea0ac25d987de1f84376405d38d43337d31cf5

ecsv:
Can you please retest with the current master.

Because there was some confusion regarding the branches: The lede-17.01 branch never had this problem - the incorrectly committed compat-hacks.h was in the master branch (the one for OpenWrt master)

SvenRoederer:
https://github.com/openwrt-routing/packages/commit/467f0b61e219c8e9d9964df604ca7fa87fc7548e already fixed the issue building openwrt-master with lede-17.01.
Our buildbots also created images with openwrt-packages master (f48f332) based on lede-17.01 (https://github.com/openwrt/openwrt/commit/dca4dfacf22e93017c90c308dbeec27a77b4f338)

PR #338, opened 2017-12-26
lede-17.01: batman-adv: Merge bugfixes from v2017.4

ecsv:
@simonwunderlich 

batctl
======

* Handle allocation error in vlan_get_link_parse
* Simplify concatenation of pathnames
* tcpdump: Fix types for for TT v1
* Free nl_sock when if_nametoindex failed
* Free nl_sock on genl_ctrl_resolve error
* Handle nl_cb_alloc errors
* Handle nl_socket_alloc errors
* Handle nlmsg_alloc errors
* Handle allocation error for path_buff
* Handle failure during hash_iterator allocation
* Print dummy value when localtime failed
* suppress implicit-fallthrough compiler warning
* change PATH_BUFF_LEN to maximal possible value

batman-adv
==========

* Fix lock for ogm cnt access in batadv_iv_ogm_calc_tq
* Fix check of retrieved orig_gw in batadv_v_gw_is_eligible
* Always initialize fragment header priority
* Avoid spurious warnings from bat_v neigh_cmp implementation
* fix TT sync flag inconsistencies
* Accept only filled wifi station info
* Use default throughput value on cfg80211 error
* do not add loop detection mac addresses to global tt
* Fix rx packet/bytes stats on local ARP reply
* handle race condition for claims between gateways
* fix memory leak when dropping packet from other gateway
* drop unicast packets from other backbone gw
* prevent duplication of ARP replies when DAT is used
* prevent multiple ARP replies sent by gateways if dat enabled
* Initialize gw sel_class via batadv_algo
* Keep fragments equally sized
* decrease maximum fragment size
* don't add loop detect macs to TT
* Treat NET_XMIT_CN as transmit successfully


p4u:
Does this upgrade fix #334 ?

ecsv:
@p4u, No, this is branch (lede-17.01) doesn't even have the code mentioned in #334. You are searching for the PR #337 (master)

PR #340, opened 2017-12-31
bird: update to 2.0.0

Noltari:
Merges bird4 & bird6 into a single package

Signed-off-by: Álvaro Fernández Rojas <noltari@gmail.com>


p4u:
Is this tested and compatible with the current bird-openwrt package?

https://github.com/openwrt-routing/packages/tree/master/bird-openwrt/bird4-openwrt

eloicaso:
No, it isn't.

I'll test it soon and see if I need to tie the current version with bird1.6.3 or if the naming change (from bird4/bird6 to just bird) will block the upgrade. Anyway, I've already started looking into this to adopt it. Take into account that bird2.0.0 seems to be unstable in some OSs (e.g. segfaults) according to bird's mail list.

tohojo:
This should probably be bumped to 2.0.2 - there are quite a few bugfixes, including several crash bugs, which is what I think @eloicaso is referring to from the mailing list :)

mwarning:
Superceded by #378

PR #343, opened 2018-01-05
nodogsplash2: Fix Startup Failure

bluewavenet:
This change will fix a problem where NoDogSplash2 Fails to start if the device it binds to is not ready.

### DIFF
--- a/nodogsplash2/files/nodogsplash.config
+++ b/nodogsplash2/files/nodogsplash.config
@@ -12,14 +12,14 @@ config nodogsplash
   # Use plain configuration file
   #option config '/etc/nodogsplash/nodogsplash.conf'
 
-  # Set the network interface the users are connected to
+  # Use this option to set the network interface the users are connected to
   # Must not be used with option gatewayinterface
   # This option automatically identifies the active lan device for nodogsplash to bind to
   # This option may fail if the device configured for this interface is not up when nodogsplash starts at boot time
   # You may change this to any valid virtual lan interface that has been defined, eg lan, lan2, public_lan wlan2 etc
   # option network 'lan'
   
-  # Set the device nogogsplash will bind to
+  # Use this option to set the device nogogsplash will bind to
   # Must not be used with option network
   # The nodogsplash init script will wait for this device to be up before loading the nodogsplash service
   # You may change this to any valid lan device eg br-lan, wlan0, eth0.1 etc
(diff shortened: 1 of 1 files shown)


bluewavenet:
@mwarning 
This came to light whilst chasing mwan3 conflict. Has nothing to do with mwan3 issue though.

mwarning:
Every commit needs to be prefixed with the package name (the package name is usually lower case).
Every pull request needs to increase the package revision PKG_RELEASE.
Every commit should have a Signed-off-by line (e.g. see https://github.com/openwrt-routing/packages/pull/344/commits/0caaf6dde03ce9ee2fa67e3d980afec553bbec19)

This repository is less strict with rules as compared to openwrt/packages, but it good style.

Please also keep the commit messages short unless it is a very important and difficult topic.

bluewavenet:
@mwarning 
Thanks for the suggestions - a lot to learn and hard to find time to do so.
I have manually added a signed off by line as I cannot see any other way in github.com.
Are you saying I should have also incremented PKG_RELEASE in Makefile as you have in #344 ?

mwarning:
#344 is a release update. In this case PKG_RELEASE is set back to 1 (if it is not already 1).
The point is that every package can be uniquely identified by PKG_VERSION and PKG_RELEASE.

mwarning:
Sorry for missing to answer you actual question. We need to update PKG_RELEASE in this pull request.

bluewavenet:
@mwarning 
I understand, I think. So normally my pull request would contain my changes PLUS an update to Makefile?
In this case, as it is a new release and we are both updating something at the same time, unless I am missing something, then:
a. if mine is merged first, then yours second, everything will be as it should be with neither of us having to do anything else.
or
b. Yours is merged first, then I do a new pull request incorporating my changes, then mine needs to be merged to complete the business.

mwarning:
Yes. In case of b, you can update you pull request. But a new pull request might be easier.

bluewavenet:
Ok now I know :-)
In this case are we going with a. or b. ? 
Case b. will result in version 2.0.1-2 and that kind of upsets the balance of the Universe inside my head :-D

mwarning:
You will survive. :P

Anyway, if we do not need option network, we should get rid of it.

bluewavenet:
Option network is useful for some hardware so not a problem leaving it in, but it should not be the default.

bluewavenet:
Back to my question.
Case a: or case b: ?
Case a: is two clicks for you.
Case b: is more work and will send us straight to NDS 2.0.1-2 with -1 never existing.
I will respect your decision :-D

PR #345, opened 2018-01-08
bird: remove incorrect build depends

NeoRaider:
Build depends must refer to source rather than binary package names
(readline and ncurses instead of libreadline and libncurses).

As libreadline and libncurses are also runtime dependencies of the
individual binary packages, build depends on the corresponding source
packages are implied and the incorrect build depends can simply be removed.

See also: https://github.com/openwrt/packages/pull/5370

### DIFF
--- a/bird/Makefile
+++ b/bird/Makefile
@@ -1,4 +1,4 @@
-# 
+#
 # Copyright (C) 2009-2016 OpenWrt.org
 #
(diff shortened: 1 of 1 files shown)


NeoRaider:
@Noltari

NeoRaider:
Has been fixed in c5f18d5a2a83cb1109aa68ce52dfb05f4c307a4f.

PR #350, opened 2018-02-27
lede-17.01: batman-adv: Merge bugfixes from 2018.0

ecsv:
@simonwunderlich

batman-adv
==========

* fix packet checksum in receive path
* invalidate checksum on fragment reassembly
* Ignore invalid batadv_iv_gw during netlink send
* Ignore invalid batadv_v_gw during netlink send
* Fix netlink dumping of BLA claims
* Fix netlink dumping of BLA backbones
* Fix internal interface indices types

### DIFF
--- a/batman-adv/patches/0024-batman-adv-fix-packet-checksum-in-receive-path.patch
+++ b/batman-adv/patches/0024-batman-adv-fix-packet-checksum-in-receive-path.patch
@@ -1,0 +1,43 @@
+From: Matthias Schiffer <mschiffer@universe-factory.net>
+Date: Tue, 23 Jan 2018 10:59:49 +0100
+Subject: [PATCH] batman-adv: fix packet checksum in receive path
+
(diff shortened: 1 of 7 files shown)


SvenRoederer:
Can you bump the version of the package? 
I'm not sure about the best naming-choice, probably
> PKG_VERSION:=2016.5-2018.0?

ecsv:
No, it is still 2016.5 with extra fixes. It is still using the same tarball and the patches are still only OpenWrt only changes.

SvenRoederer:
So changing Makefiles ```PKG_RELEASE``` from 5 to 6 should be done, to indicate the update? That it's "seen" by ```opkg update```

ecsv:
Ok, this is correct. I missed that again.

SvenRoederer:
Will you create a PR for this?

ecsv:
Would be nice when you could do that.

SvenRoederer:
> simonwunderlich merged commit 190f1b0

PR #352, opened 2018-03-01
bmx6/7: add pkg mirror hash

dangowrt:
@axn please add PKG_MIRROR_HASH when using larger git repos to reduce traffic and load on the buildbots.


dangowrt:
@axn also noticed that you didn't reset PKG_RELEASE to 1 when bumping the upstream sources. It's too late for now, but please do so in future to avoid confusion.

PR #353, opened 2018-03-15
cjdns: store node sources to openwrt download dir

Dimazhan:


mwarning:
The patch changes indentation from white spaces to tabs. This results in the target file having different indentations. Please fix.

@lgierth can you also have a look? I do not know cjdns.

wfleurant:
I'm not 110% certain this is a good idea. In the end, cjdns will fetch the nodejs binary (not sources).  
Any changes should be sent upstream to cjdelisle/cjdns (to avoid maintaining patches).  
What's the reasoning behind moving it out of it's build directory?  

Anyway, I'd rather we:  
- append a variable to the Build/Compile section
- update node.sh
  -  download the nodejs sources to $DL_DIR without renaming destination file 
  -  prevent extracting nodejs to $DL_DIR, extract to build or (perhaps) staging

Then, we could just add this:  

```
define Build/Compile
	....
	NODEDIR="$(DL_DIR)" \
	NODEEXT="$(PKG_BUILD_DIR)" \
	exec ./do
endef

PR #357, opened 2018-03-27
nogogsplash2: Add compatibility with mwan3 v2

bluewavenet:
nogogsplash2: Add compatibility with mwan3 version 2
Changes in mwan3 version 2 resulted in previous recommended fix for Nodogsplash to fail.
Setting fw_mark to use bits set "higher" than used in mwan3 solves the problem. This works also for all previous versions of Nodogsplash.
Author-name: Rob White
Maintainer: Moritz Warning <moritzwarning@web.de>
Compile not required
Run tested: Run and tested on LEDE 170104
Description: fw_mark settings compatible with mwan3 version 2

### DIFF
--- a/nodogsplash2/files/nodogsplash.config
+++ b/nodogsplash2/files/nodogsplash.config
@@ -73,6 +73,6 @@ config nodogsplash
   #list trustedmac '00:00:C0:01:D0:1D'
 
   # Set FW_MARK for compatibilty with other OpenWrt Packages eg mwan3, sqm etc.
-  list fw_mark_authenticated '520'
-  list fw_mark_trusted '510'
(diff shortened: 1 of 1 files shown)


bluewavenet:
@mwarning 
There is a new release of mwan3 today where the mmx_mask default is changed but also configurable in config so now both mwan3 and NDS are compatible by default AND compatible with SQM. We are good to go with this PR :-)

bluewavenet:
@mwarning 
I only have access via the Github web ui at the moment so quicker if I close this and redo ;-)

mwarning:
ok. (Otherwise use "git commit --amend" and "git push -f" to change the commit message)

PR #358, opened 2018-03-30
 nogogsplash2: Add compatibility with mwan3 v2

bluewavenet:
nogogsplash2: Add compatibility with mwan3 version 2
Changes in mwan3 version 2 resulted in previous recommended fix for Nodogsplash to fail.
Setting fw_mark to use bits set "higher" than used in mwan3 solves the problem. This works also for all previous versions of Nodogsplash.
Author-name: Rob White


mwarning:
ok thanks!

bluewavenet:
Do'h ;-D

PR #362, opened 2018-04-15
bmx7: keep bmx7 secret keys on sysupgrade

aparcar:
Sysupgrade currently leads to a loss of the content of /etc/bmx7 which contains the secret keys.
After the reboot the device creates new IPs leading to a new "identity" aka ID, shortID and IPv6 changes.


aparcar:
@axn @dangowrt ping

PR #360, opened 2018-04-08
bmx7: convert init script to use procd

dangowrt:
Using procd allows better control and monitoring of the bmx7 process.
Keep polling for mac80211.sh process for now until we find a better
way to make sure that wireless configuration has been generated before
starting bmx7.
Setting maximum core size to 20000 looks dangerous on small devices,
but keep it as it was before, we shall discuss this separately.

Signed-off-by: Daniel Golle <daniel@makrotopia.org>

PR #363, opened 2018-04-16
bmx7: configReload on initd reload

aparcar:
Instead of restarting bmx7 completely and loosing all connections.

@axn


aparcar:
this should be integrated in #360 so this PR is not needed

PR #364, opened 2018-04-24
bmx7: use configReaload on service reload

aparcar:
It's #363 for procdbmx7: use configReaload on service reload

@axn ping

Signed-off-by: Paul Spooren <spooren@informatik.uni-leipzig.de>


axn:
Btw, i thunk sendink a kill -HUB signal to bmx7 pid (e.g. $(cat /var/run/bmx7/pid) ) could be used instead of 'bmx7 - configReloaD'. On the other hand, not sure, but could be that the return value of 'bmx7 -c configReload || /etc/init.d/bmx7 restart' could be used to force a hard restart in case some parameters (like plugins) could not be reloaded on the fly

aparcar:
Your choice, what do you think is best?

axn:
Did you test if -c configReload returns a non-zero exit-code in case the new config can not be applied on-the fly? If yes i'd go for your current approach

aparcar:
Yes the return value is non zero if the config is broken. However your choice now results in "gently reload bmx7 config or if it fails crash bmx7 and thereby stop it". Meaning, shouldn't the `/etc/init.d/bmx7 reload`  be *safe* so that it doen't crash bmx7 if you screwed the config?

axn:
Hmm, good argument! I do not see an easy way to figure out if a failed on-the-fly reload would succeed with a restart. A visible warning may help to creste awareness. But ok, probably doi g only a soft configReload is the better way.

aparcar:
so be it. Please merge

PR #365, opened 2018-04-25
nodogsplash2: Add NDS Restart Hook for Firewall

bluewavenet:
nodogsplash2: Add NDS Restart Hook for Firewall
NodogSplash iptables entries are overwritten if the OpenWrt Firewall is restarted.
This change adds a restart hook to firewall.user to restart NodogSplash if Firewall is restarted.
Author-name: Rob White

### DIFF
--- a/nodogsplash2/files/nodogsplash.init
+++ b/nodogsplash2/files/nodogsplash.init
@@ -248,8 +248,22 @@ create_instance() {
   procd_close_instance
 }
 
+depends() {
+  if [ "$1" = "iptables" ] ; then
+  if $WD_DIR/ndsctl status > /dev/null; then
+  echo " * Restarting NodogSplash" 
+  /etc/init.d/nodogsplash restart
+  fi
+  fi
+}
+
 start_service() {
   include /lib/functions
+
+  if [ "$(grep -c "if \[ \-f \'/etc/init.d/nodogsplash\' \] ; " /etc/firewall.user)" = "0" ] ; then
+  echo "Installing NDS hook for iptables" 2>&1 | logger
+  printf "\n  if [ -f '/etc/init.d/nodogsplash' ] ; then /etc/init.d/nodogsplash depends iptables ; fi\n" >> /etc/firewall.user
+  fi
 
   mkdir -p /tmp/etc/
   config_load nodogsplash


bluewavenet:
@mwarning 
This provides the NDS restart hook discussed in NodogSplash issue 144
https://github.com/nodogsplash/nodogsplash/issues/141

jow-:
Please do not rewrite `/etc/firewall.user`, this is a user configuration file. Take a look at the `bcp38` or `miniupnpd` instead to see how to properly hook into the firewall.

bluewavenet:
@jow- 
What I am trying to do is trigger a NodogSplash restart when the firewall is updated.
Looking at your example in bcp38, unless I am missing something, that is to do the reverse, ie trigger a firewall restart if bcp38 is updated.

jow-:
It does both. Look at how the firewall include script is registered in postinstall.

nodogsplash2/files/nodogsplash.init
```diff
@@ -248,9 +248,23 @@ create_instance() {
   procd_close_instance
 }
 
+depends() {
+  if [ "$1" = "iptables" ] ; then
+  if $WD_DIR/ndsctl status > /dev/null; then
```
mwarning: some indentation would be nice here

mwarning:
@bluewavenet
We would need something like this in nodogsplash2/Makefile
```
define Package/nodogsplash2/install
  $(INSTALL_DIR) $(1)/etc/uci-defaults
  $(INSTALL_BIN) ./files/nodogsplash2.defaults $(1)/etc/uci-defaults/nodogsplash2
  ...
```
And have a script `nodogsplash2/files/nodogsplash2.defaults`. It is run only during the package installation.

But I wonder if there isn't a better solution using procd hooks.

nodogsplash2/files/nodogsplash.init
bluewavenet: @mwarning 
I made it look like the rest of the script :-D
I'll put my indentation back in.

bluewavenet:
@jow- 
something like:
service_triggers()
{
	procd_add_config_trigger "config.change" "firewall" /etc/init.d/nodogsplash restart
}

Does not seem to work. Restarting firewall, even without any actual changes to the firewall config, overwrites the NDS iptables entries.

jow-:
No, see https://github.com/openwrt/packages/blob/master/net/bcp38/Makefile#L57 and https://github.com/openwrt/packages/blob/master/net/bcp38/files/bcp38.defaults

bluewavenet:
@mwarning 
As it stands at the moment, when NDS starts, it checks to see if the hook is installed. If not it installs it.
If the hook is installed and firewall is restarted, then the hook calls /etc/init.d/nodogsplash depends iptables.
NDS init.d script then checks to see if NDS is running and if it is, restarts it and if not running then does nothing.

mwarning:
@jow- ah, now I understand. An include section in the firewall config is used to run a script when the firewall is restarted.

jow-:
Yes, instead of modifying firewall.user (which is not invoked on reloads, only on restarts btw) add a new, ideally named, include section (to avoid aggregating duplicates on package reinstalls) and set it to type=script, reload=1, path=/path/to/restart/helper.sh

jow-:
If only `fw3 restart` is clearing the rules, but not `fw3 reload`, then leave out reload=1 in the include section.

bluewavenet:
The problem is finding a way of making firewall tell nodogsplash that it (firewall) has just restarted.
What I have proposed here does just that. In addition nodogsplash will check the hook is in place every time it is started.
It adds a single line to firewall.user, the purpose of which is to add custom firewall commands.
I have had this running in intensive production environments for over eight months so works well for me. Unfortunately I do not have time right now to figure out how, or even if, service_triggers can be made to do this although it does look interesting.

jow-:
This isn't about service triggers. You're supposed to add a further `config include` section to `/etc/config/firewall` which calls a helepr script restarting nodogsplash.

In-place modifications of user configurations are not acceptable for packages.

bluewavenet:
@jow- 
" add a new, ideally named, include section (to avoid aggregating duplicates on package reinstalls) and set it to type=script, reload=1, path=/path/to/restart/helper.sh"
Where?
Duplicate entries into firewall.user should never happen as NDS will check before adding.

bluewavenet:
@jow- 
Writing to /etc/config/firewall is acceptible?

bluewavenet:
@jow- 
@mwarning 
I am testing a firewall uci config version now. Will close this one and do a new PR when tests complete.

PR #369, opened 2018-04-26
nodogsplash2: Add NDS Restart Hook for Firewall

bluewavenet:
nodogsplash2: Add NDS Restart Hook for Firewall
NodogSplash iptables entries are overwritten if the OpenWrt Firewall is restarted.
This change adds an include hook to /etc/config/firewall to restart NodogSplash if Firewall is restarted.
Author-name: Rob White
Signed-off-by: Rob White rob@blue-wave.net

### DIFF
--- a/nodogsplash2/files/nodogsplash.init
+++ b/nodogsplash2/files/nodogsplash.init
@@ -267,9 +267,9 @@ start_service() {
   
   if [ $(uci get nodogsplash.@nodogsplash[0].fwhook_enabled) = "1" ] ; then
     if ! uci get firewall.nodogsplash2.path &> /dev/null ; then
-        if [ -f '/tmp/etc/ndshook.include' ] ; then
-          rm /tmp/etc/ndshook.include
-        fi
+      if [ -f '/tmp/etc/ndshook.include' ] ; then
+        rm /tmp/etc/ndshook.include
+      fi
       uci delete firewall.nodogsplash2 2> /dev/null
       uci set firewall.nodogsplash2=include
       uci set firewall.nodogsplash2.type=script
--- a/nodogsplash2/files/nodogsplash.init
+++ b/nodogsplash2/files/nodogsplash.init
@@ -252,10 +252,11 @@ depends() {
   if [ "$1" = "iptables" ] ; then
     if [ $(uci get nodogsplash.@nodogsplash[0].fwhook_enabled) = "1" ] ; then
       if $WD_DIR/ndsctl status > /dev/null; then
-      echo " * Restarting NodogSplash" 
-      /etc/init.d/nodogsplash restart
+       echo " * Restarting NodogSplash" 
+       /etc/init.d/nodogsplash restart
       fi
-    else   echo " * NodogSplash fwhook is disabled" 
+    else   echo " * NodogSplash fwhook is disabled"
+    
     fi
   fi
 }
--- a/nodogsplash2/files/nodogsplash.init
+++ b/nodogsplash2/files/nodogsplash.init
@@ -251,6 +251,26 @@ create_instance() {
 start_service() {
   include /lib/functions
 
+  if ! uci get firewall.nodogsplash2.path &> /dev/null ; then
+      if [ -f '/etc/nodogsplash/ndshook.sh' ] ; then
+        rm /etc/nodogsplash/ndshook.sh
+      fi
+    uci delete firewall.nodogsplash2 2> /dev/null
+    uci set firewall.nodogsplash2=include
+    uci set firewall.nodogsplash2.type=script
+    uci set firewall.nodogsplash2.path='/etc/nodogsplash/ndshook.sh'
+    uci commit firewall
+    /etc/init.d/firewall restart  2>&1 | logger
+  fi
+
+  if [ ! -f '/etc/nodogsplash/ndshook.sh' ] ; then
+    printf "#NDS Restart Hook Created by Nodogsplash. Do not modify.\n" > /etc/nodogsplash/ndshook.sh
+    printf "  if /usr/bin/ndsctl status > /dev/null; then\n" >> /etc/nodogsplash/ndshook.sh
+    printf "    echo \" * Restarting NodogSplash \"\n" >> /etc/nodogsplash/ndshook.sh
+    printf "    /etc/init.d/nodogsplash restart\n  fi\n" >> /etc/nodogsplash/ndshook.sh
+    chmod +x /etc/nodogsplash/ndshook.sh
+  fi
+
   mkdir -p /tmp/etc/
   config_load nodogsplash
 
--- a/nodogsplash2/files/nodogsplash.init
+++ b/nodogsplash2/files/nodogsplash.init
@@ -255,8 +255,8 @@ depends() {
        echo " * Restarting NodogSplash" 
        /etc/init.d/nodogsplash restart
       fi
-    else   echo " * NodogSplash fwhook is disabled"
-    
+    else
+      echo " * NodogSplash fwhook is disabled"
     fi
   fi
 }
--- a/nodogsplash2/files/nodogsplash.init
+++ b/nodogsplash2/files/nodogsplash.init
@@ -248,30 +248,41 @@ create_instance() {
   procd_close_instance
 }
 
+depends() {
+  if [ "$1" = "iptables" ] ; then
+    if [ $(uci get nodogsplash.@nodogsplash[0].fwhook_enabled) = "1" ] ; then
+      if $WD_DIR/ndsctl status > /dev/null; then
+      echo " * Restarting NodogSplash" 
+      /etc/init.d/nodogsplash restart
+      fi
+    else   echo " * NodogSplash fwhook is disabled" 
+    fi
+  fi
+}
+
 start_service() {
   include /lib/functions
-
-  if ! uci get firewall.nodogsplash2.path &> /dev/null ; then
-      if [ -f '/etc/nodogsplash/ndshook.sh' ] ; then
-        rm /etc/nodogsplash/ndshook.sh
-      fi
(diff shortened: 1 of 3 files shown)


bluewavenet:
@mwarning 
@jow- 
In this version the nodogsplash init.d script checks for the existence of the restart hook script and the include in the firewall uci config and creates them if not present.
The include is removed if nodogsplash is uninstalled.

bluewavenet:
Annoyingly the 2 commits have the wrong description and as I am using the github web ui I cannot edit them. This PR does have the correct description though.
If this is an issue I will close and reissue.

nodogsplash2/files/nodogsplash.init
```diff
@@ -251,6 +251,26 @@ create_instance() {
...
+    uci set firewall.nodogsplash2.path='/etc/nodogsplash/ndshook.sh'
+    uci commit firewall
+    /etc/init.d/firewall restart  2>&1 | logger
+  fi
+
+  if [ ! -f '/etc/nodogsplash/ndshook.sh' ] ; then
+    printf "#NDS Restart Hook Created by Nodogsplash. Do not modify.\n" > /etc/nodogsplash/ndshook.sh
+    printf "  if /usr/bin/ndsctl status > /dev/null; then\n" >> /etc/nodogsplash/ndshook.sh
+    printf "    echo \" * Restarting NodogSplash \"\n" >> /etc/nodogsplash/ndshook.sh
+    printf "    /etc/init.d/nodogsplash restart\n  fi\n" >> /etc/nodogsplash/ndshook.sh
```
mwarning: This might be more elegant instead all of the printf calls:
```
	cat <<-EOF > /etc/nodogsplash/ndshook.sh
		#NDS Restart Hook Created by Nodogsplash. Do not modify.
		if /usr/bin/ndsctl status > /dev/null; then
		  echo " * Restarting NodogSplash"
		  /etc/init.d/nodogsplash restart
		fi
	EOF
```

mwarning:
Anyway, I think the suggested way from #365 with a script in /etc/uci-defaults is still the preferred way.
I would agree that your approach is easier to understand, but it's good to keep conventions.

nodogsplash2/files/nodogsplash.init
bluewavenet: Beauty is in the eye of the beholder!
I prefer the symmetry of the printf lines. lol

bluewavenet:
The convention to be kept, referred to in #365, is to **refrain from modifying user config files** from a running package ie the file firewall.user in that case.

The contents of /etc/uci-defaults are deleted on first boot and only relevant for custom images and are "run once".

We need a "self healing" hook that fixes up changes that might be made when people are making firewall additions/deletions after NDS is installed.

Here we have **both** the NDS restart hook script called on firewall reset (residing in /etc/nodogsplash) and the Firewall UCI Include, being checked every time NDS starts and created/recreated if required to keep the NDS reset hook functional when NDS is running.

In addition, un-installing NDS will clean up the Firewall UCI config using the post remove, postrm script defined in the make file.

nodogsplash2/files/nodogsplash.init
mwarning: I think it is better to put the script in a file that is installed instead of creating it programmatically. Less noise and it would be easier for the file system to compress.

mwarning:
Anyway, self healing leads to fighting the user. If they delete a setting, then it is the users responsibilty. my2cents

bluewavenet:
Nodogsplash init.d already writes the NDS conf file into /tmp/etc
Would it be more acceptable for the hook script to be written there?

This is not intended to fight the user, but intended to fix the security issue of NDS failing if the firewall is restarted (usually after adding or deleting some rules) and fixing the issue as a part of the normal startup of NDS. 
https://github.com/nodogsplash/nodogsplash/issues/141

I have a number of production systems that are remotely administered and adding rules for remote access to lan installed devices eg access points, heating controllers, etc is quite common and it is essential that NDS keeps running regardless of any edits to firewall config, otherwise it is a security issue if NDS fails. 

A common quick update for firewall rules (by on site support) is often to copy a firewall uci config file and restart - despite my attempts to stop them from doing this.
For this reason I implemented the firewall.user version on our firmware.

The method presented here is better, keeps NDS running and keeps to the convention mentioned above.

mwarning:
imho, /tmp/etc is somewhat better if the script is created dynamically.

What I do not understand is why the script has to be created at run time, instead of installing ndshook.sh as part of the package installation/update.
As far as I understand it, you want to prevent users from deleting/modifying ndshook.sh. But this is not a configuration file, and if the user messes with system scripts (hook or init scripts alike), then they keep the pieces if something breaks, so to speak.

bluewavenet:
I understand your thinking. However this is really an integral part of the startup of NDS where the init.d script sets up the environment for NDS to run. In my initial PR here it was integral to the init.d script and the hook was called by firewall.user.
This version uses an include in the firewall configuration as a way round not using the firewall.user file and the consequence of this is the requirement for an additional script file to actually include. (lets call it an include file).
What we really need is a mechanism for advanced users to disable/enable the hook with the include file part of the /tmp/etc environment. 
This is very simple to do with the addition of a line in the /etc/config uci file. Something like:
option fwhook_enabled '1'
then checking in the init.d script.
I will make this change, test, and update here.
With an enhancement to, say, ndsctl, we could reinitialise the NDS iptables settings without restarting.
If this was available the hook would then be very client friendly as well as system friendly.

jow-:
Its not my call to make as @mwarning is maintaining the package but I do not really like the auto-generated script file thing at all.

1) it escapes proper file tracking by opkg (opkg remove nodogsplash will not nuke it unless additional, redundant steps are implemented in postrm)
2) due to 1 it is also not properly updated on "opkg upgrade" calls unless you add yet more logic to either postinst or the init script to detect outdated versions and update them accordingly
3) `/etc/` is polluted with package components. I know that there are some system code files in `/etc/`, e.g. `rc.common` but those reside there due to historical reason, we should not further expand such misused

I do understand that the hook file is an integral part, but so is `/etc/rc.common` or `/lib/functions.sh` - yet these file do not receive such protection.

I will look into expanding fw3 in the meanwhile to not require a script at all, the include can then specify a command line directly, without the need for an intermediate script.

bluewavenet:
@jow- 
I understand and agree with your concerns. I am proposing to move the include file to the NDS created /tmp/etc/nodogsplash folder and as a result it will not survive a reboot.
This mitigates all three of your concerns, I think.

I like the idea of specifying a command in the firewall config. It gets my vote!

bluewavenet:
@mwarning 
Absolutely aware, things never stand still.
The first change would be if @jow- has time for the addition of the firewall "direct command" config option!

nodogsplash2/files/nodogsplash.init
```diff
@@ -248,10 +248,41 @@ create_instance() {
   procd_close_instance
 }
 
+depends() {
+  if [ "$1" = "iptables" ] ; then
+    if [ $(uci get nodogsplash.@nodogsplash[0].fwhook_enabled) = "1" ] ; then
+      if $WD_DIR/ndsctl status > /dev/null; then
+      echo " * Restarting NodogSplash"
```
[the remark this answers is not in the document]
mwarning: Also, even when NDS is disabled, the "Restarting NodogSplash" message will still be emitted.
Maybe leave the message out. Or if necessary, check the return value of `/etc/init.d/nodogsplash restart`.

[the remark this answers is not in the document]
bluewavenet: ok ;-)

nodogsplash2/files/nodogsplash.init
```diff
@@ -248,10 +248,41 @@ create_instance() {
...
 }
 
+depends() {
+  if [ "$1" = "iptables" ] ; then
+    if [ $(uci get nodogsplash.@nodogsplash[0].fwhook_enabled) = "1" ] ; then
+      if $WD_DIR/ndsctl status > /dev/null; then
+      echo " * Restarting NodogSplash" 
+      /etc/init.d/nodogsplash restart
+      fi
+    else   echo " * NodogSplash fwhook is disabled"
```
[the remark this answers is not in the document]
bluewavenet: done

bluewavenet:
@mwarning 
Ready, I think :-)

nodogsplash2/files/nodogsplash.init
```diff
@@ -248,10 +248,42 @@ create_instance() {
...
+depends() {
+  if [ "$1" = "iptables" ] ; then
+    if [ $(uci get nodogsplash.@nodogsplash[0].fwhook_enabled) = "1" ] ; then
+      if $WD_DIR/ndsctl status > /dev/null; then
+       echo " * Restarting NodogSplash" 
+       /etc/init.d/nodogsplash restart
+      fi
+    else   echo " * NodogSplash fwhook is disabled"
+    
+    fi
```
mwarning: Instead of this:
```
  else   echo " * NodogSplash fwhook is disabled"

  fi
```
do
```
  else
    echo " * NodogSplash fwhook is disabled"
  fi
```

bluewavenet: done

nodogsplash2/files/nodogsplash.init
```diff
@@ -248,10 +248,42 @@ create_instance() {
...
 start_service() {
   include /lib/functions
-
   mkdir -p /tmp/etc/
+  
+  if [ $(uci get nodogsplash.@nodogsplash[0].fwhook_enabled) = "1" ] ; then
+    if ! uci get firewall.nodogsplash2.path &> /dev/null ; then
+        if [ -f '/tmp/etc/ndshook.include' ] ; then
+          rm /tmp/etc/ndshook.include
+        fi
```
mwarning: The last indentation is four spaces here. But it should be two.
I think that should be all.

Issue #277, opened 2017-02-21
Remove or keep bird-openwrt packets?


p4u:
The bird integration to UCI/LUCI was made by Eloi, a student which is now out of the map. I pushed the packets to the oprnwrt-routing repository with the idea that he would maintain them, however I don't think it is gonna happen. Is there anyone interested on taking care of the maintenance? If not, I would remove them at least from the official repository.  

https://github.com/openwrt-routing/packages/tree/master/bird-openwrt

eloicaso:
Dear Pau,

As you say, I have not been actively maintaining the package (lack of time). However, I just have started updating it as I could make it part of my MSc. project.

tohojo:
Any plans to support Bird 2.0?

eloicaso:
Yes, I already started looking at it and how Bird2 manages configuration now. However, I have had no spare time yet to properly start developing it. Moreover, I am still unsure about how bird1/2 separation will be handled (https://github.com/openwrt-routing/packages/pull/340)

tohojo:
Eloi <notifications@github.com> writes:

> Yes, I already started looking at it and how Bird2 manages
> configuration now. However, I have had no spare time yet to properly
> start developing it. Moreover, I am still unsure about how bird1/2
> separation will be handled
> (https://github.com/openwrt-routing/packages/pull/340)

Ah, awesome, didn't see that pull request. Not sure there's a reason to
keep bird 1.6 around if you can support the 2.0 config format in uci.
The upgrade cost is mostly in configuration file format changes...

briantopping:
Please don't delete this package. Even if it is out of date, it is better than nothing.

rogerpueyo:
Hi @p4u,

I'm maintaining the bird1-openwrt packages now, so you may want to close this issue.

Cheers!

PR #375, opened 2018-05-08
bird: Update to v2.0.2

tohojo:
Add a couple of backported patches that fixes the build version + a couple
of fixes to the Babel implementation. Also update the init file to use procd.

Signed-off-by: Toke Høiland-Jørgensen <toke@toke.dk>

This is based on, and supersedes, https://github.com/openwrt-routing/packages/pull/340

This just replaced the old 1.6 version with 2.0. The stability issues should be fixed with 2.0.2 + the backported patches included here. There is the issue of config compatibility, which could be a reason to keep the old version around, but if that is a concern I'd suggest keeping a compatibility package for bird 1.6 and having 2.0 be the 'main' version...


tohojo:
Cc @Noltari  @eloicaso @p4u

mwarning:
Do you also get this warning message?
```
[mwarning@xanax openwrt]$ make defconfig
WARNING: Makefile 'package/feeds/routing/bird4-openwrt/Makefile' has a dependency on 'bird4', which does not exist
WARNING: Makefile 'package/feeds/routing/bird6-openwrt/Makefile' has a dependency on 'bird6', which does not exist
```

tohojo:
Ah, yeah, bird-openwrt only works with bird1. See https://github.com/openwrt-routing/packages/issues/277

I guess that a decision needs to be made on whether to keep that and/or update it. Maybe it is better to package both versions of Bird for now, then... I'll things to allow that and open a new pull request.

mwarning:
Maybe rename the package to bird2.

(the existence of bird4 and bird6 are a bit silly)

tohojo:
I did, in the updated pull request: https://github.com/openwrt-routing/packages/pull/378

eloicaso:
Already commented on #378, moving to bird2 makes sense as the ipv4/6 bird instances are just options and not part of the binary name.

I agree that once bird-openwrt is updated to use the new v2 features it will be much nicer to use avoiding the old bird4/6 calls and double packaging.

I'll keep you updated when I have something worth sharing.

PR #378, opened 2018-05-16
Add a package for Bird 2.0

tohojo:
This adds a new package for bird 2.0, renaming the old Bird packages to allow them to co-exist (as bird1 and bird2 + suffixes).

This is based on, and supercedes, https://github.com/openwrt-routing/packages/pull/340

@Noltari: I added myself as maintainer for bird2 as I wasn't sure whether you'd want to maintain both packages.

Cc @eloicaso @p4u


tohojo:
Yeah, I realise that; but I want to avoid the bird2/4/6 confusion. I did add CONFLICTS to all the new packages, so installing the new ones should get rid of the old ones.

I guess I could add stub packages for the old names that pull in the new ones instead? Not sure that works with opkg, though?

Also, this would only affect people who run 'opkg update' on master builds; how common is that, and how many hoops do we want to jump through to accommodate it? :)

mwarning:
Ah, I did not notice the CONFLICTS line. :-)
If the bird package is not upgradeable, then bird1 for the updated package should be fine.

eloicaso:
I'm happy with the changes proposed to the bird-openwrt itself. It will make it easy to maintain and the conflict in bird2 will avoid adding odd restrictions in the bird-openwrt package (I started playing with the  few flags available to prevent a package to go beyond version X.Y.Z and it wasn't working at all). I'm hoping to start implementing the bird2-openwrt package on june~july, so I'll keep an eye on any updates on bird2 :)

Cheers!

mwarning:
It builds for me. Let me know when it is ready to be merged.

tohojo:
As far as I'm concerned you can go ahead and merge :)

tohojo:
Awesome, thanks! :)

PR #390, opened 2018-06-13
alred: use procd fix #388

### LINKED ISSUE #388: alfread: use procd
https://github.com/openwrt-routing/packages/blob/013dcc6817eac52d7684667dccac0c27b7607ecf/alfred/files/alfred.init#L10

Looks like Alfread doesn't use procd? If there is no reason except a missing PR, I'd create one.

aparcar:
Currently I have no testbed where I could check if it works as excepted, will do soon if nobody else finds the time.

Signed-off-by: Paul Spooren <mail@aparcar.org>


aparcar:
should solve https://github.com/openwrt-routing/packages/issues/388

nicopace:
@axn guess this is relevant for you (don't know if you are around here)

aparcar:
@mwarning @ecsv

alfred/Makefile
```diff
@@ -76,7 +76,7 @@ define Package/alfred/install
 	[ "x$(CONFIG_PACKAGE_ALFRED_VIS)" == "xy" ] && cp -fpR  $(PKG_BUILD_DIR)/vis/batadv-vis $(1)/usr/sbin/ ; true
 	[ "x$(CONFIG_PACKAGE_ALFRED_GPSD)" == "xy" ] && cp -fpR  $(PKG_BUILD_DIR)/gpsd/alfred-gpsd $(1)/usr/sbin/ ; true
 	$(INSTALL_DIR) $(1)/etc/init.d
-	$(INSTALL_BIN) ./files/alfred.init $(1)/etc/init.d/alfred
+	$(INSTALL_BIN) ./files/alfred.procd $(1)/etc/init.d/alfred
```
ecsv: Either I am blind or no one else is using this *.procd extension in OpenWrt. If you want to move/rename something then use the complete paths (as seen on the device) in alfred/files/. But this should most likely be placed in a separate commit

alfred/files/alfred.procd
```diff
@@ -103,41 +94,38 @@ alfred_start()
...
-
-	if [ "$vis_enable" = "1" ]; then
-		echo "${initscript}: starting batadv-vis"
-		SERVICE_PID_FILE="$pid_file_vis"
-		service_start /usr/sbin/batadv-vis ${vis_args}
-	fi
-
-	if [ "$run_facters" = "1" ]; then
+	procd_open_instance "alfred"
+	echo "starting alfred"
```
ecsv: Why this echo? Other procd scripts also don't do this.

alfred/Makefile
aparcar: Okay, renamed back to `alfred.init`

alfred/files/alfred.init
```diff
@@ -103,41 +94,38 @@ alfred_start()
...
-
-	if [ "$vis_enable" = "1" ]; then
-		echo "${initscript}: starting batadv-vis"
-		SERVICE_PID_FILE="$pid_file_vis"
-		service_start /usr/sbin/batadv-vis ${vis_args}
-	fi
-
-	if [ "$run_facters" = "1" ]; then
+	procd_open_instance "alfred"
+	echo "starting alfred"
```
ecsv: Why this echo? Other procd scripts also don't do this.

alfred/files/alfred.procd
aparcar: The echo was there [before](https://github.com/openwrt-routing/packages/pull/390/files#diff-9cac72d591240485e6d7dd48f4aaa90eL115) and I though of sticking as much to the "original" init file as possible

alfred/files/alfred.init
aparcar: @ecsv so if you like I'd remove that?

ecsv: I just find it odd that it is there. I am neither the maintainer nor have I write access to the repository.

mwarning: @aparcar please remove it. Thanks.

aparcar: this here as well? https://github.com/openwrt-routing/packages/pull/390/files/a75b388cea308b8867b2d45a6ab39be89c4847f8#diff-9cac72d591240485e6d7dd48f4aaa90eR108

alfred/files/alfred.init
```diff
@@ -103,41 +94,38 @@ alfred_start()
...
+	procd_set_param command /usr/sbin/alfred
+	procd_append_param command ${alfred_args}
+	procd_close_instance
+
+	[ "$vis_enable" == "1" ] && {
+		procd_open_instance "batadv-vis"
+		echo "starting batadv-vis"
```
aparcar: @mwarning @ecsv remove this as well?

mwarning: Imho, yes. Only output errors.

aparcar: done

aparcar:
anything else to change?

ecsv:
Ddid anyone test that it actually works? At least this is what the PR currently says:

> Currently I have no testbed where I could check if it works as excepted, will do soon if nobody else finds the time.

alfred/files/alfred.init
```diff
@@ -103,41 +94,36 @@ alfred_start()
...
-		service_start /usr/sbin/batadv-vis ${vis_args}
-	fi
-
-	if [ "$run_facters" = "1" ]; then
+	procd_open_instance "alfred"
+	procd_set_param command /usr/sbin/alfred
+	procd_append_param command ${alfred_args}
+	procd_close_instance
+
+	[ "$vis_enable" == "1" ] && {
```
mwarning: == is bash syntax, use =

alfred/files/alfred.init
```diff
@@ -103,41 +94,36 @@ alfred_start()
...
 }
 
-stop()
-{
-	SERVICE_PID_FILE="$pid_file_alfred"
-	service_stop /usr/sbin/alfred
-	SERVICE_PID_FILE="$pid_file_vis"
-	[ -x /usr/sbin/batadv-vis ] && service_stop /usr/sbin/batadv-vis
+stop_service() {
 	sed "\|for file in $facters_dir/\* ; do |d" -i /etc/crontabs/root
```
mwarning: I get an error here because /etc/crontabs/root does not exist. Maybe add 2> /dev/null or check if the file exists?

mwarning:
I can't test right now. I just managed to break both my test devices. :]

aparcar:
@mwarning requested changes applied

mwarning:
@aparcar this needs to be tested, otherwise it can't be merged.

adrianschmutzler:
What kind of testing do you require? Is a plain "alfred works in my configuration" enough or do you aim at specific tests?

mwarning:
@adrianschmutzler  "alfred works in my configuration" is enough.

mwarning:
@aparcar anyway, what happend to the vis code?

adrianschmutzler:
Is the $enabled variable still used? I just see it set, but not evaluated ...

adrianschmutzler:
Successfully tested in a configuration with
        option mode 'slave'
        option batmanif 'none'
        option start_vis '0'
        option run_facters '0'
(I used the master from 2018-07-19, so I removed the Makefile part of the patch. The rest applies ...)

Still would be interested in what happened to the enable var ...

alfred/files/alfred.init
```diff
@@ -103,41 +94,38 @@ alfred_start()
 	return 0
 }
 
-start()
-{
+start_service() {
 	config_load "alfred"
 	config_foreach alfred_start alfred
-
-	if [ "$enable" = "0" ]; then
```
mwarning: I do not think this can be removed. It is set in the original file.

adrianschmutzler: Just had a closer look:
"config_foreach alfred_start alfred" (L109) runs alfred_start for each alfred uci config entry.
After alfred_start went through successfully, enable is set to 1 (L84).
With the if clause, we make sure that alfred is not started if:
- no alfred uci config is present
- a config which is present does not go through setup successfully.

So, we "need" the "if [ "$enable" = "0" ]; then exit 0 fi" except alfred is supposed to start in any case (i.e. without config).

alfred/Makefile
```diff
@@ -8,10 +8,9 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=alfred
-PKG_VERSION:=2018.1
-PKG_RELEASE:=1
-PKG_HASH:=808fa6acf65c7a8e26405115176a5587157f746108cbe5dd974788eb05416d76
-
+PKG_VERSION:=2018.2
+PKG_RELEASE:=0
+PKG_HASH:=2bbe9c4fc19a6529c40f7c872406fad8070d145ede970806bd2194e07fba8018
```
adrianschmutzler: Maybe squash this into the first patch and add the enable-if-clause ...

aparcar: done https://github.com/openwrt-routing/packages/pull/390/files#diff-9cac72d591240485e6d7dd48f4aaa90eR101

adrianschmutzler:
LGTM

aparcar:
thanks for the reviews and merge

mwarning:
@ecsv aynway, are you the effective maintainer of the package? The makefile lists no maintainer.

ecsv:
No, I only propose new package versions because I try to submit the new batman-adv releases in time.

adrianschmutzler:
The latest change with {exit 0} seems to be wrong syntax and broke alfred start for me. Please have a look at: https://github.com/openwrt-routing/packages/pull/397

PR #393, opened 2018-07-02
luci-app-cjdns: bump cjdns and update luci admin

wfleurant:
- cjdns: bump to crashey @ a88859364
- luci-app-cjdns: adds layer 2 and 3 address display column
- luci-app-cjdns: fixes switch label pinger
- luci-app-cjdns: get version from addr string
- luci-app-cjdns: parse addr string for peerStats publicKey


mwarning:
Is this a backport? Release branches should only contain backports or bug fixes.
Also, has a88859364 a crash problem? That is how I read it.

@lgierth maybe you can take a look. :-)

wfleurant:
- no, no crash problem. branch "crashey" is development. 
- no, not a backport  
- features and newer versions are put directly to master, no?

wfleurant:
moved to PR #394

PR #395, opened 2018-07-16
nodogsplash2: cleanup

mwarning:
Changes:
* rework restart hook to use recommended install procedures
* fix multiple small bugs and style issues


mwarning:
@bluewavenet can you take a look if the changes have broken something or should be changed?

Following changes might cause issues:
* option gatewayinterface and network behavior (why do we have an extra option network?)
* restart hook changes

bluewavenet:
@mwarning 
Re: option network - This appeared with version 2 I think, no idea why. I think it is more useful in Debian version?

bluewavenet:
Re: Restart hook
The new hook script checks if nodogsplash is enabled and if it is, issues a restart.
This is bad in the scenario that NDS has been stopped while other config changes are being carried out. Restarting firewall will then start NDS when not intended.
The old script checks if nodogsplash is actually running and if it is, restarts it, otherwise does nothing.

bluewavenet:
Re: Rework
If I recall correctly the issue was not (eventually) install procedures but avoiding writing to /etc at runtime. The final incarnation of the old version avoided this. Also if the firewall config is rewritten for other reasons, this new version will be broken. Yes, as discussed previously, this is the problem of the person rewriting the firewall config. Personally I prefer resilience against inadvertent changes, but as long as it works, all well and good.

bluewavenet:
On the subject of package cleanups, should we drop version 1 from 18?

mwarning:
@bluewavenet that might be a good idea by now. Since we don't maintain it anymore.

mwarning:
@bluewavenet the new hook scripts does check if nds is running. It uses `ndsctl status`. Or do I miss something?

bluewavenet:
I saw your first version of restart.sh and made the comment but you had already changed it :-D

mwarning:
@bluewavenet ah :-)

mwarning:
Ok, I have tested the changes now. Looks good.
Additional changes:
- removed nolog, errors on boot are logged anyway
- removed module tests. Were inactive since nds 1.x anyway.

mwarning:
@bluewavenet I do not think we should/can fight a user messing with foreign firewall settings. So that approach should imo suffice.

bluewavenet:
It all looks good. I'm happy.  If you are too then ready to merge?
I have some new PR's coming with additional options so will be good to have this merged first.

PR #397, opened 2018-07-21
alfred: Fix alfred.init

adrianschmutzler:
The commit 45db0e6 introduced curly brackets which are not correct
syntax and thus broke the script.

At the moment, alfred is not starting!


adrianschmutzler:
@mwarning @ecsv

mwarning:
@aparcar don't forget testing the final changes :->

PR #401, opened 2018-08-08
nodogsplash2: update to 3.0 and merge packages

mwarning:
Nodogsplash 3.0 was released.

The code has been successfully  tested with current openwrt master on wt3020.
Package nodogsplash2 is not needed anymore. The main feature from 1.0.0 that had been missing from 2.0.0 has been added in 3.0.0.


mwarning:
@bluewavenet

mwarning:
- the files are package specific and should be part of the package
  - the debian package is an exception, since we have no official debian package
- a package maintainer usually has no access the application repository

bluewavenet:
I'm happy for both  v1 and v2 to be replaced by v3 AND backport to 17.01

mwarning:
@bluewavenet a backport sounds good, but that needs a different pull request.

PR #411, opened 2018-09-15
nodogsplash: add missing config options

bluewavenet:
sessiontimeout (minutes) until client is forced out
checkinterval (seconds) interval at which nodogsplash checks client timeout status

Add real world useful values to timeout options.

Signed-off-by: Rob White rob@blue-wave.net


bluewavenet:
@mwarning

PR #418, opened 2018-09-27
babeld: Update to 1.8.3

robimarko:
Update to recently released version 1.8.3.
This contains some nice fixes for kernels 4.16 and up which will become interesting as OpenWrt moves to 4.19
Full changelog:
  * Fixed a read-only two byte buffer overflow in the packet parser.
    This is a read-only overflow, and hence most probably not exploitable.
  * Fixed an issue with creating unreachable routes on recent kernels
    (4.16 and up).  Thanks to Christof Schulze.
  * Notice interface changes faster by listening to more netlink events.
    Thanks to Christof Schulze.
  * Fixed a local interface issue when an interface has no link-local
address. Thanks to Christof Schulze.

Signed-off-by: Robert Marko <robimarko@gmail.com>

### DIFF
--- a/babeld/Makefile
+++ b/babeld/Makefile
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=babeld
-PKG_VERSION:=1.8.2
+PKG_VERSION:=1.8.3
 PKG_RELEASE:=1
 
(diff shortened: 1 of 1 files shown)


robimarko:
Didnt see that there was one already made for this

PR #419, opened 2018-09-28
nodogsplash: fix download link

mwarning:
The tagging of the release was wrong, as a result, the hash is different from the usual package tar.gz.
The content has been verified to be the same.

Signed-off-by: Moritz Warning <moritzwarning@web.de>


mwarning:
@bluewavenet

mwarning:
@bluewavenet removing the v would be something we might forget next time.

PR #437, opened 2018-12-09
luci-app-bmx6: Avoid race condition in bmx6json.lua get()

rogerpueyo:
This commit adds a dirty fix to issue https://github.com/openwrt-routing/packages/issues/436. Just adding the sys.exec("") call makes the topology and the graph be shown properly, not even a sleep command is needed.

Ideally, the wget() function should be reviewed by any of the contributors (@nicopace, @p4u) to properly identify where the problem is and fix it.

Signed-off-by: Roger Pueyo Centelles <roger.pueyo@guifi.net>


mwarning:
@p4u ping :)

p4u:
I was delaying this merge to get a deeper look on this issue, but got no time. So for now I'll accept this patch solution. Thanks @rogerpueyo

PR #438, opened 2018-12-13
hnetd: add compatiblity with openssl 1.1.x

cotequeiroz:
Tested with openwrt-mipsel_74kc with both openssl 1.1.1a and 1.0.2q.
Patch submitted upstream at sbyx/hnetd#66

Signed-off-by: Eneas U de Queiroz <cote2004-github@yahoo.com>


mwarning:
thanks

PR #440, opened 2019-01-14
bird2: Bump to v2.0.4

tohojo:
Bird 2.0.4 was released on March 1st.

Signed-off-by: Toke Høiland-Jørgensen <toke@redhat.com>


BKPepe:
May I ask you, if you try to compile it? I have tried to compile your PR, but end up with the following errors as it seems it doesn't compile.

Just tested it on Turris MOX, OpenWrt master:

```
Applying ./patches/0001-BGP-Fix-extended-next-hop-handling.patch using plaintext: 
patching file proto/bgp/attrs.c
Hunk #1 FAILED at 295.
1 out of 1 hunk FAILED -- saving rejects to file proto/bgp/attrs.c.rej
Patch failed!  Please fix ./patches/0001-BGP-Fix-extended-next-hop-handling.patch!
```
and so on. I'm not sure if the patch is needed. So, I remove it. Also, I checked these patches are not necessary as they're already included in bird.
0002-Filter-Added-missing-instruction-comparators.patch
0003-babel-Fix-type-of-route-entry-router-ID.patch
0004-Babel-Add-option-to-randomize-router-ID.patch
0005-Better-initialization-of-random-generator.patch
0008-Filter-Add-support-for-src-filter-op-to-access-SADR-.patch

Not sure about 0007-Fix-build-version.patch as it might be included in bird, too.

So, I suggest my changes as with them I was able to compile it on Turris MOX. https://github.com/BKPepe/openwrt-routing/commit/22c295e9ec41a07785c0e2d4996afe964b927b0c
However, I haven't tested it for now and will get back to you, once I do that.

tohojo:
Ah, right, yeah, those are all backports that are now included upstream. Totally forgot about those; thanks for pointing it out. Will fix! :)

mikma:
I get a build error in babel without the following modification:

```
diff --git a/bird2/patches/0001-babel-Set-onlink-flag-for-IPv4-routes-with-unreachab.patch b/bird2/patches/0001-babel-Set-onlink-flag-for-IPv4-routes-with-unreachab.patch
index 528051d..5d194d2 100644
--- a/bird2/patches/0001-babel-Set-onlink-flag-for-IPv4-routes-with-unreachab.patch
+++ b/bird2/patches/0001-babel-Set-onlink-flag-for-IPv4-routes-with-unreachab.patch
@@ -29,7 +29,7 @@ index ce05191c..afd2eb19 100644
 +     * makes it possible to, e.g., assign /32 addresses on a mesh interface and
 +     * have routing work.
 +     */
-+    if (!neigh_find2(&p->p, &r->next_hop, r->neigh->ifa->iface, 0))
++    if (!neigh_find(&p->p, r->next_hop, r->neigh->ifa->iface, 0))
 +          a0.nh.flags = RNF_ONLINK;
 +
      rta *a = rta_lookup(&a0);

```

tohojo:
Ah yes, so it would seem... updated :)

BKPepe:
There's already version 2.0.4.

tohojo:
Josef Schlehofer <notifications@github.com> writes:

> There's already version 2.0.4.

Ah, doh! Seems I've been a little out of touch with upstream lately...
Will update this later, thanks! :)

n8v8R:
@tohojo it would seem that Master is still serving 2.0.2 and the push did not make it for the past 4 months? Let alone 2.0.4 available from source?

tohojo:
Argh, seems I totally forgot about this, so sorry! Pushed an update for 2.0.4 now.

Unfortunately, I don't have permissions to merge. @jow- care to do the honours? :)

PR #448, opened 2019-02-18
batctl: fix wrong binary-name for batctl-default

SvenRoederer:
* correct binary for batctl-default to /usr/libexec/batctl-default; it was ```/usr/sbin/batctl -> /usr/libexec/batctl-full```
* while at it fix description of batctl-full


adrianschmutzler:
LGTM and fixed the described issue in my OpenWrt-derived firmware.

adrianschmutzler:
@ecsv @simonwunderlich

ecsv:
Looks good, thanks

PR #456, opened 2019-03-14
nodogsplash: fix invalid pointer bug when clock is turned back

mwarning:
Fixes a invalid pointer error when the clock is set into the past while nodogsplash is running.

Tested with current OpenWrt master and Nexx wt3020.


mwarning:
@bluewavenet

PR #457, opened 2019-03-16
nodogsplash: Bump to v3.3.0-1

bluewavenet:
Bump to v3.3.0-1


bluewavenet:
@mwarning

nodogsplash/Makefile
```diff
@@ -41,7 +41,6 @@ define Package/nodogsplash/description
 endef
 
 define Package/nodogsplash/install
-	$(CP) ./files/* $(1)/
```
mwarning: This is the preferred way to place files. Let's keep it.

mwarning: Nvm, you use the files from the source package

bluewavenet: Yes, everything critical to the package is in the package source with no files here to place.

mwarning: It feels odd that we would then have a different openwrt package folder in nodogsplash/openwrt/ compared to openwrt-routing/nodogsplash. I meant this to be a copy of the oprnwrt package.

@bluewavenet I would like to propose to add the updated files for now and to postpone the descision for the next release.

bluewavenet: @mwarning 
I created nodogsplash/openwrt with quality control in mind and the intention of safely automating all aspects of an OpenWrt release (apart from the version and checksum).

The current way of doing things is very prone to errors and omissions and is subject to unlinked changes that may be done on either "side".

This change also brings OpenWrt release in line with the principle of the Debian release (although that is not complete either).

I would prefer to make this change now for the above reasons. If unforeseen problems arise it can be fixed one way or another, although I do not expect any issues.

mwarning: Ok, I see the advantages of having all things in one repository. But this mixes packaging and source code too much for my taste. We just end part of the OpenWrt package files here and some as part of the nodogsplash source package.
I do not see that this improves quality control. Separation of code is a cornerstone of software programming. Sure, we can give it a try, but I might have to revert that part of the pull request in the future if it causes confusion.

PR #458, opened 2019-03-19
batman-adv: Fix uci commit target

adrianschmutzler:
Although batman-adv config is cycled in config_foreach, changes
are made in the network config. Thus, this one has to be committed.

Signed-off-by: Adrian Schmutzler <freifunk@adrianschmutzler.de>


adrianschmutzler:
@ecsv

PR #459, opened 2019-03-24
[OLSRd] Bump OLSRd version to 0.9.7

pmelange:
This patch bumps OLSRd from 0.9.6.2 to 0.9.7 (released 06-March-2019)

I have tested this on the Freifunk-Berlin Backbone (250+ node, 800+ links) in a mixed environment with multiple versions of OLSRd.

Signed-off-by: pmelange <isprotejesvalkata@gmail.com>


mwarning:
@zioproto

zioproto:
@mwarning I am not actively using `olsrd` since 4 years now. We need a new maintainer :(

In the `Makefile` the variable `PKG_VERSION` looks wrong. 
It looks like this change that introduced the bug is already 3 years old 3d25d7793535aa524fd3f09a71656fea30c3b73f

`PKG_VERSION` should be an integer, that is used to track when something changes in the package but the source code we are packaging is unchanged. For example a change in the init scripts is tracked increasing `PKG_VERSION`.

Maybe @bittorf wants to take ownership on the olsrd package ?
I dont feel to be the right person anymore to follow up on this.

thanks !

Saverio

mwarning:
@zioproto Should we remove the maintainer line? What is the procedure here?

bittorf:
i will take ownership of the package and will update it regarding the PKG_VERSION

PR #461, opened 2019-03-28
batman-adv 2019.1

ecsv:
batman-adv 2019.1
=================

* support latest kernels (3.16 - 5.1)
* implement meshif, hardif and vlan config interface via netlink
* snoop DHCP ACKs to improve DAT cache efficiency
* bugs squashed:
  - avoid potential reference counter underflows and use-after free errors for several objects
  - fix GCC warning when B.A.T.M.A.N. V is build in but cfg80211 is not available
  - reject too small VLAN packets before they are processed further

batctl 2019.1
=============

* introduce support for batadv meshif, hardif and vlan configuration via netlink
* replace multicast_mode with multicast_forceflood settings subcommand
* add hop_penalty setting subcommand

alfred 2019.1
=============

* synchronization of batman-adv netlink header


batctl/Makefile
```diff
@@ -122,6 +122,7 @@ config-n := \
 	log \
 	loglevel \
 	mcast_flags \
+	multicast_forceflood \
 	multicast_mode \
```
adrianschmutzler: Comment says:

> replace multicast_mode with multicast_forceflood settings subcommand

Here, you ADD and do not replace. Haven't checked the code. Is this intentional?

adrianschmutzler:
@ecsv Note my comment above. Don't know whether you get notified automatically for code annotations...

ecsv:
Yes, this is expected. We don't want to break backward compatibility yet. You will not see the old command in the help but it is still accepted.

batctl/Makefile
ecsv: Yes, this is expected. We don't want to break backward compatibility yet. You will not see the old command in the help but it is still accepted.

The PR itself has just the changelogs from the release.

PR #477, opened 2019-06-06
Fix babeld interface configuration in init script

G10h4ck:
Without this PR the interface configuration lines are duplicated for each option, probably causing the options to not be properly applied to each interface. This PR commit solve the issue.

### DIFF
--- a/babeld/files/babeld.init
+++ b/babeld/files/babeld.init
@@ -177,6 +177,10 @@ babel_configpaths() {
 start_service() {
 	mkdir -p /var/lib
 	mkdir -p /var/etc
+	mkdir -p "$OTHERCONFIGDIR"
+
+	# Start by emptying the generated config file
+	>"$CONFIGFILE"
 
 	# First load the whole config file, without callbacks, so that we are
 	# aware of all "ignore" options in the second pass.  This also allows
@@ -186,14 +190,6 @@ start_service() {
 	# Configure alternative configuration file and directory
 	config_foreach babel_configpaths "general"
 
-	# Start by emptying the generated config file
-	>"$CONFIGFILE"
-	# Import dynamic config files
-	mkdir -p "$OTHERCONFIGDIR"
-	for f in "$OTHERCONFIGDIR"/*.conf; do
-		[ -f "$f" ] && cat "$f" >> "$CONFIGFILE"
-	done
-
 	# Parse general and interface sections thanks to the "config_cb()"
 	# callback.  This allows to loop over all options without having to
 	# know their name in advance.
@@ -201,13 +197,24 @@ start_service() {
 	config_load babeld
 	# Parse filters separately, since we know which options we expect
 	config_foreach babel_filter filter
+
+	# Import dynamic config files
+	for f in "$OTHERCONFIGDIR"/*.conf; do
+		[ -f "$f" ] && cat "$f" >> "$CONFIGFILE"
+	done
+
 	procd_open_instance
 	# Using multiple config files is supported since babeld 1.5.1
 	procd_set_param command /usr/sbin/babeld -I "" -c "$OTHERCONFIGFILE" -c "$CONFIGFILE"
 	procd_set_param stdout 1
 	procd_set_param stderr 1
 	procd_set_param file "$OTHERCONFIGFILE" "$OTHERCONFIGDIR"/*.conf "$CONFIGFILE"
-	procd_set_param respawn
+
+	### Respawn automatically when process dies, after waiting respawn_timeout seconds
+	### If respawn_retry consecutives respawns die before respawn_threshold seconds (i.e. they crash)
+	### it will stop trying and leave it dead.
+	procd_set_param respawn ${respawn_threshold:-60} ${respawn_timeout:-3} ${respawn_retry:-5}
+
 	procd_close_instance
 }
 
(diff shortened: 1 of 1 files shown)


babeld/files/babeld.init
```diff
@@ -113,34 +113,39 @@ babel_config_cb() {
 		}
 	;;
 	"interface")
+	{
+		local _ignored
+		config_get_bool _ignored "$section" 'ignore' 0
+		[ "$_ignored" == "1" ] && return
```
adrianschmutzler: If you are moving this around anyways, maybe change to single equal sign? File has /bin/sh shebang and other comparisons in this file are also using single = ...

babeld/files/babeld.init
```diff
@@ -113,34 +113,39 @@ babel_config_cb() {
...
 			local option="$1"
 			local value="$2"
-			local _interface
+
 			# "option ifname" is a special option, don't actually
 			# generate configuration for it.
-			[ "$option" = "ifname" ] && return
-			[ -n "$interface" ] && _interface="interface $interface" || _interface="default"
-			cfg_append "$_interface ${option//_/-} $value"
+			[ "$option" == "ifname" ] && return
```
adrianschmutzler: Same here, consider single equal sign

babeld/files/babeld.init
adrianschmutzler: Is this an intended behavior change:
Previously, option_cb() was set to do nothing if interface was ignored.
Now, option_cb() is not set at all, since you return here...

babeld/files/babeld.init
```diff
@@ -113,34 +113,39 @@ babel_config_cb() {
...
-		unset interface
+		local interface
 		network_get_device interface "$_ifname" || interface="$_ifname"
-		option_cb() {
+
+		# Ignore section if ifname is not provided
+		[ -z "$interface" ] &&
+		{
+			logger -s -t babeld "Ignoring interface section which miss ifname"
+			return
```
adrianschmutzler: Like above, if interface is empty/not set, option_cb() won't exist. Is this intended?

babeld/files/babeld.init
```diff
@@ -133,7 +133,12 @@ babel_config_cb() {
 		}
 
 		interfaceFile="${OTHERCONFIGDIR}/interface_$interface.conf"
-		echo "interface $interface" > "$interfaceFile"
+
+		if [ "$interface" == "default" ]; then
```
adrianschmutzler: Single = here, too.

G10h4ck:
Inside ash tests `=` and `==` are equivalent, I do prefer `==` because is much more familiar to programmers of other languages too...

babeld/files/babeld.init
G10h4ck: so should I define `option_cb() { return; }` to make sure nothing is done?

adrianschmutzler:
> Inside ash tests = and == are equivalent, I do prefer == because is much more familiar to programmers of other languages too...

Well, previous style in file was = and /bin/sh shebang would also require =

But it is not my decision what will be accepted at the end ...

babeld/files/babeld.init
adrianschmutzler: If I understand the original code correctly, it would be 
```
[ "$_ignored" == "1" ] && 
{
option_cb() { return; }
return
}
```
which is terribly ugly and may be the case why it was solved with the other order initially.

However, I not familiar with this script and just stumbled over this issue when reading the PR. So my initial comment was more a question than a statement.

zorun:
@G10h4ck can you be more specific about the bug you encountered? Do you have an example UCI config that triggers the bug, and the corresponding babeld config that is generated?

Also, it looks like you are introducing new per-interface config files in your patch. Can you explain why it's needed?

G10h4ck:
As an example this config snipped
```
config interface 'lm_net_eth1_2_babeld_if'
        option ifname 'wlan1-mesh_17'
        option wired 'false'
        option faraway true
```

Would generate this config snippet

```
interface wlan1-mesh_17
interface wlan1-mesh_17 wired false
interface wlan1-mesh_17 faraway true
```

Which one will then babeld consider? From our tests we weren't able to conclude that all the options were correctly applied.

About using different files it is because as it is not always clear in which moment opntion_cb is called  to use the already existent /tmp/babel.d/*.conf seemed the most readable/elegant way of doing it.

zorun:
Sorry, I haven't had time to look at this in details yet (organisation of the Battlemesh takes lots of time..)

If you're right, it's a pretty huge bug, I'm surprised nobody came across this before!

zorun:
@G10h4ck so, I finally had a look at this issue!

However, I cannot reproduce the underlying bug: even when splitting the configuration over several lines, like in the following example, I verified that all configuration items are correctly applied by babeld:

```
interface wlan0 
interface wlan0 type wireless
interface wlan0 enable-timestamps true
interface wlan0 hello-interval 10
interface wlan0 rxcost 150
```

As you can see in https://github.com/jech/babeld/blob/master/configuration.c#L711 , babeld explicitly merges configuration structures when it encounters a new "interface XXX" line for an existing interface.  If for some reasons it doesn't work as expected, this is a bug in babeld and it should be reported.

One possible reason you got confused about some options not applying: `wired true/false` is deprecated, and I'm not sure what it does nowadays. You should use `type wired`, `type wireless` or `type tunnel`, see the man page.  I also wrote a somewhat detailed documentation for OpenWrt: https://openwrt.org/docs/guide-user/services/babeld

Given all this, I don't like the extra complexity and brittleness of generating files in `/tmp/` if it's not necessary, so I am going to decline your PR...

zorun:
Btw, you were telling me you also had some issues with the "status" command?

I just tested with the following config:
- OpenWrt 18.06.1
- Babeld 1.8.2-1

and it does print babeld output in `logread` when I run `/etc/init.d/babeld status`

Which version were you using?

G10h4ck:
I expected it to be printed on the stdout, in this case feel free to decline the PR ;)

PR #478, opened 2019-06-15
batman-adv: Add support for B.A.T.M.A.N. V hardif settings

ecsv:
batctl currently supports settings which are either mesh interface or vlan specific. But B.A.T.M.A.N. V introduced two additional settings which are hard (slave) interface specific.

* elp_interval
* throughput_override

To support these, an additional command prefix called hardif is implemented for some sysfs commands:

    $ batctl -m bat0 hardif eth0 ..

The batadv_hardif proto has to expose them to allow automatic configuration. The default configuration would be:

    # add eth0 as hardif to bat0
    config interface 'bat0_hardif_eth0'
    	option proto 'batadv_hardif'
    	option master 'bat0'
    	option ifname 'eth0'
    	option mtu '1536'
    	option 'elp_interval' 500
    	# string like '1mbit' is accepted instead of kbit for throughput_override
    	option 'throughput_override' '0'

----

Please try to do a a fast-forward merge to master.

    git clone -b master git@github.com:openwrt-routing/packages.git openwrt-routing
    git -C openwrt-routing merge --ff-only origin/openwrt-19.07
    git -C openwrt-routing push origin master

If it is not possible anymore than I will just create a new pull request for it.


simonwunderlich:
the fast forward merge seemed to have worked too.

adrianschmutzler:
@ecsv Off-topic, but: Is there an automated way to create patches as you did here: Without first From: line, without [PATCH] and without the version at the end?

ecsv:
@adrianschmutzler This is similar to what gluon does with `make update-patches`. But I am just using an alias in my .gitconfig:

    [pretty]
    	fixes = Fixes: %h (\"%s\")
    [alias]
    	format-simple = format-patch --no-stat --full-index --no-renames --binary --diff-algorithm=histogram --no-signature --format=format:'From: %an <%ae>%nDate: %aD%nSubject: %B'
    	fixes = log -1 --abbrev=12 --pretty=fixes


Then it is possible to just do something like:

    git format-simple -1

adrianschmutzler:
Thanks, that's exactly what I needed.

PR #479, opened 2019-06-17
Build fixes

neheb:
https://downloads.openwrt.org/snapshots/faillogs/arc_archs/routing/

### DIFF
--- a/mcproxy/patches/0003-add-sourcefilter.patch
+++ b/mcproxy/patches/0003-add-sourcefilter.patch
@@ -4,7 +4,7 @@
  #include <numeric>
  #include <unistd.h>
  
-+#ifndef __GLIBC__
++#if !defined(__GLIBC__) || defined(__UCLIBC__)
 +#include "sourcefilter.cpp"
 +#endif /* __GLIBC__ */
 +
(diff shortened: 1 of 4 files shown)


ecsv:
@jow- @mwarning @lgierth @dangowrt 
@stepet92 @Noltari @sbyx

mcproxy/patches/0003-add-sourcefilter.patch
```diff
@@ -4,7 +4,7 @@
  #include <numeric>
  #include <unistd.h>
  
-+#ifndef __GLIBC__
++#if !defined(__GLIBC__) || defined(__UCLIBC__)
```
jow-: Is this really supposed to mean "not glibc or uclibc"? Or did you rather mean `#if  !defined(__GLIBC__) && !defined(__UCLIBC__)` ?

cotequeiroz: I haven't tested this, but reading the commit message it looks correct.  
uClibc-ng defines `__GLIBC__` but needs to include `sourcefilter.cpp`.

neheb: It's correct. It's needed for uclibc-ng but not glibc.

neheb:
Any other problems?

PR #485, opened 2019-07-22
batman-adv: Use Linux 5.2 API again

ecsv:
The mac80211 package version was updated to a 5.2 based one in commit 0b2c42ced21a ("mac80211: Update to version 5.2-rc7"). The batman-adv module must therefore no longer the pre-5.2 API.


SvenRoederer:
this should fix the build-error from the OpenWrt-devel ML (http://lists.infradead.org/pipermail/openwrt-devel/2019-July/018064.html)

PR #486, opened 2019-07-23
nodogsplash: Backport Version 4.0.0.

bluewavenet:
nodogsplash: Backport Version 4.0.0.

This version has numerous fixes and enhancements.
It is compatible with the previous v3.2.1 release and onwards.

Important fixes in this version:
  * Fix memory handling bug, issue https://github.com/nodogsplash/nodogsplash/issues/341 [mwarning] [stevo01]
  * ndsctl_thread - ignore interupts when returning from epoll [lynxis]
  * auth.c - use correct types to prevent cast and comparement of uint and int [lynxis]
  * openwrt/init.d - prevent start of the daemon if configuration generation fails [lynxis]
  * Generate Error 403 Forbidden, if Gateway Port is accessed directly [bluewavenet]
  * Validate fasremoteip as a valid dotted format IPv4 address  [bluewavenet] [mwarning]
  * Prevent client CPD  "Too Many Redirects" error.  [bluewavenet]


bluewavenet:
@mwarning

bluewavenet:
@mwarning 
Do we need to do the same for the 19.07 branch or will it be updated before release?

mwarning:
It won't be updated before the release. We need a PR for NDS.

bluewavenet:
@mwarning 
The build system does not seem to be running for this branch - we are still stuck at the old version....

PR #487, opened 2019-07-25
cjdns-tests: Don't build on ARC

neheb:
Not supported.

https://downloads.openwrt.org/snapshots/faillogs/arc_archs/routing/cjdns/compile.txt

### DIFF
--- a/cjdns/Makefile
+++ b/cjdns/Makefile
@@ -56,7 +56,7 @@ define Package/cjdns-tests
 	TITLE:=cjdns test cases
 	URL:=https://github.com/cjdelisle/cjdns
 	MAINTAINER:=Lars Gierth <larsg@systemli.org>
-	DEPENDS:=+libpthread +librt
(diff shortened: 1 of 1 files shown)


wfleurant:
I am looking for the abiName value. Is it possible to compile on the target device? If so there is a directory `./cjdns/build_linux/dependencies/cnacl` from which you could run:  
`cc -o /dev/null -c okcompilers/abiname_xcompile.c`  

To make a plan file we'll need to match abiname output similar to this:
```
okcompilers/abiname_xcompile.c:5:9: error: #error ABI_IS_amd64__
        #error ABI_IS_amd64__
         ^~~~~
```

Do you have the above target hardware to test on? I'll see if we can fix ARC support. No worries if else.

neheb:
I'd rather just disable it so that the buildbots don't even try building. ARC is a fairly new platform. Not widely used either.

PR #489, opened 2019-07-27
batmand/vis: Convert to procd based init scripts

ecsv:
The legacy init script had various problems in comparison with procd based init scripts. It wasn't able to correctly track the running process instance and thus could:

* accidentally kill another (non init controlled) daemon instance when stop is used
* not restart the daemon depending on config changes when reload is used
* not automatically start/restart daemon when the used netdev was created/recreated

The information about a running instance and its parameters can now be handled by a global controller (procd). The process must not fork anymore and leave the control to procd. The process with its parameters can then be used by procd to trigger the stop/start of the process at the right time.

----

@elektra42


elektra42:
Thank you very much!

PR #490, opened 2019-07-28
olsrd: fix build against gpsd-3.19

RussellSenior:
The olsrd package requires gpsd as a build dependency, even if no gpsd-requiring modules are selected. This patch allows olsrd to build against gpsd-3.19, where gps_read() added two function arguments. 

Signed-off-by: Russell Senior <russell@personaltelco.net>


SvenRoederer:
There is a PR https://github.com/openwrt/packages/pull/9588 for openwrt-packages which updates to V3.19 to fix the build-errors

RussellSenior:
@fhuberts ?

fhuberts:
Looks like you've converted the commit into a patch.
I'd not for that, as I've explained before. Try to stay as close to upstream as possible please, just update your olsrd version/git commit you're building.
If you need a new olsrd release for that, then I can make it, since the patch currently sit on master only.

RussellSenior:
The current OpenWrt Makefile is set up for releases, rather than by git commit hash, so a new release would be easier for me.

fhuberts:
release v0.9.8 is now available. Have fun.

RussellSenior:
Tested on ath79 (buffalo wzr600dhp)

RussellSenior:
@zioproto  look okay?

zioproto:
@RussellSenior why you introduced `PKG_HASH` ? My understanding is that the code is checked out via git. Is this correct ?

zioproto:
In commit 3d25d7793535aa524fd3f09a71656fea30c3b73f (dated back 2016) this package is not downloading anymore `tar.gz` but it is cloning from git. No need for `PKG_HASH` then.

RussellSenior:
@zioproto removed the PKG_HASH in the latest force-push.

zioproto:
Thank you !

thuehn:
The gpsd 3.19 dependency of olsrd does not compile on an OS X Darwin machine.
 Beside the need to fix gpsd, how about changing the olsrd Maekfile to make the pud plugin and its dependency to gpsd conditional, so that it is only build the case it is selected ?

fhuberts:
OSX is not a supported platform, patches welcome.

PR #494, opened 2019-07-30
nodogsplash: Backport Version 4.0.1.

bluewavenet:
nodogsplash: Backport Version 4.0.1.

This version has numerous fixes and enhancements.
It is compatible with the previous v3.2.1 release and onwards.

Important fixes in this version:
 * Fix coding error in fas-aes.php [bluewavenet]
 * Make debuglevel platform independent [mwarning]
 * Fix memory handling bug, issue nodogsplash/nodogsplash#341 [mwarning] [stevo01]
 * ndsctl_thread - ignore interupts when returning from epoll [lynxis]
 * auth.c - use correct types to prevent cast and comparement of uint and int [lynxis]
 * openwrt/init.d - prevent start of the daemon if configuration generation fails [lynxis]
 * Generate Error 403 Forbidden, if Gateway Port is accessed directly [bluewavenet]
 * Validate fasremoteip as a valid dotted format IPv4 address [bluewavenet] [mwarning]
 * Prevent client CPD "Too Many Redirects" error. [bluewavenet]

### DIFF
--- a/nodogsplash/Makefile
+++ b/nodogsplash/Makefile
@@ -7,12 +7,12 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=nodogsplash
 PKG_FIXUP:=autoreconf
-PKG_VERSION:=3.3.2
+PKG_VERSION:=4.0.1
 PKG_RELEASE:=1
 
 PKG_SOURCE_URL:=https://codeload.github.com/nodogsplash/nodogsplash/tar.gz/v$(PKG_VERSION)?
 PKG_SOURCE:=nodogsplash-$(PKG_VERSION).tar.gz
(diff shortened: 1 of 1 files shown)


bluewavenet:
@mwarning

mwarning:
we should minimize backports to minor releases only in the future. Otherwise there would be no point in having release branches.

bluewavenet:
Indeed,I absolutely agree. The two branches were so very far behind and 19.07 is not even released yet.
Now at least both are at a good point to be frozen. 
By the way, the binaries are not getting built in 18.06 or 19.07 at the moment. Do we have to do something else or will it be started again at some point?

mwarning:
I do not know.

PR #502, opened 2019-08-21
babeld: Update to 1.8.5 and slightly improve Makefile

adrianschmutzler:
This bumps babeld to 1.8.5 and adds some simple improvements from https://github.com/openwrt/packages/pull/9408

Since babeld-1.8.5 just contains two "fixes", I would like to have it backported to 19.07 (I could also send a PR adding those two fixes as patches, but I think this would be useless additional work).

I will send a bump to the recent version 1.9.1 in a separate PR, as I would like to have both update steps (1.8.5 and 1.9.1) separately in master. (I have been told that 1.9 incompatibly changes treatment of some source-specific routes, so having 1.8.5 as separate patch will make switching easier ...).

### DIFF
--- a/babeld/Makefile
+++ b/babeld/Makefile
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=babeld
-PKG_VERSION:=1.8.4
+PKG_VERSION:=1.8.5
 PKG_RELEASE:=1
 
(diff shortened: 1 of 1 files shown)


adrianschmutzler:
@kerneis @zorun

zorun:
Superseded by #503 and #506

PR #504, opened 2019-08-22
19.07: babeld: Update to version 1.8.5

adrianschmutzler:
Version 1.8.5 is a bugfix-only release, so it can be backported
directly:

  * Fixed a bug that caused confustion between learned routes and
    imported routes (thanks to Fabian Bläse).
  * Fixed a bug that prevented install filters from being evaluated
    (thanks to Killian Lufau).

### DIFF
--- a/babeld/Makefile
+++ b/babeld/Makefile
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=babeld
-PKG_VERSION:=1.8.4
+PKG_VERSION:=1.8.5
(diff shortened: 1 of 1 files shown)


zorun:
@adrianschmutzler I am closing this PR, because I like the idea of having babeld 1.9.X for the upcoming 19.07 release (implemented in #505 )

However updating to 1.8.5 is still a good idea for the 18.06 branch, since there have been quite a lot of bug fixes since 1.8.2.  Feel free to open a PR doing that, otherwise I'll do it at some point.

PR #508, opened 2019-08-28
quagga: move quagga to packages repo

lucize:
Signed-off-by: Lucian Cristian <lucian.cristian@gmail.com>

updated to latest version and moved to packages repo


mwarning:
ref https://github.com/openwrt/packages/pull/9857

BKPepe:
I'm in favor of the removal of quagga in the master branch. As it was superseded by frr. However, frr is not available in OpenWrt 19.07. So, I think it will be good if @lucize can create a pull request with anupdate of quagga against OpenWrt 19.07 here in this repository.

lucize:
@BKPepe maybe the upcoming 19.07.1 can have frr also ?

BKPepe:
If you are going to create a pull request in packages repository, it would be OK. I'd be in favor to include it there.

lucize:
@BKPepe sorry, but I fail in pushing to 19.07, is there a help about it ?

mwarning:
@lucize I have just cherry-picked the quagga 1.1.1 package from openwrt-18.06 to openwrt-19.07. IF you have another update for quagga, then please open a pull request and ping me. :-)

mwarning:
Hm, it seem like quagga is not going into packages (https://github.com/openwrt/packages/pull/9857).
Imho, quagga can stay in openwrt-routing as long it is maintained.

BKPepe:
That's was my point that it does not make sense to move it to packages based on its discussion as quagga development is stalled and it should be superseded by frr.

So, @lucize should do pull request here same as in the packages repository to update quagga in OpenWrt 19.07 and in the same time create a pull request in packages repository to backport frr from master to  OpenWrt 19.07 and by this way, we can remove quagga from `master` branch of this repository.

lucize:
@mwarning you can pick 9857 to stable
@BKPepe the problem is that I get an error if I want to push to origin/openwrt-19.07, I remember that I did this some time but now I can't, libyang needs updated also
so if you have a short copy paste example on how to push to stable I'll add it, imo frr has more options and we can remove quagga

PR #517, opened 2019-10-20
nodogsplash: Version 4.3.3 release

bluewavenet:
nodogsplash: Version 4.3.3 release

Compiled and tested on snapshot SDK mips_24kc and arm_cortex-a7_neon-vfpv4

This version fixes two issues that can cause NDS to lock or crash, one, a coding error that leads to memory corruption and two, deadlocks in iptables and ndsctl. Both of these issues occur at high loads and/or at high CPD detection rates.

In addition, in some circumstances, a deauthenticated client running a vpn may have suffered from querystring truncation causing vpn failure.

Some minor updates are also included.

Extract from changelog:

 * Fix Memory corruption at high loads [bluewavenet]
 * Prevent iptables and ndsctl deadlocks [lynxis]
 * Prevent query string truncation for deauthenticated client when client is using some types of vpn software [bluewavenet]
 * Add debuglevel logging in the case of a firewall restart in OpenWrt [bluewavenet]
 * Return error 403(forbidden) when client attempts to use a forbidden http method [bluewavenet]

### DIFF
--- a/nodogsplash/Makefile
+++ b/nodogsplash/Makefile
@@ -7,12 +7,12 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=nodogsplash
 PKG_FIXUP:=autoreconf
-PKG_VERSION:=4.3.0
+PKG_VERSION:=4.3.3
 PKG_RELEASE:=1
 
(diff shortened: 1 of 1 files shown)


bluewavenet:
@mwarning

PR #522, opened 2019-10-31
ahcpd: Replace PKG_MD5SUM with PKG_HASH

neheb:
PKG_MD5SUM is deprecated and does not accurately represent the SHA256 hash.


BKPepe:
This is wrong!

`WARNING: PKG_HASH uses deprecated hash, set to a4622e817d2b2a9b878653f085585bd57f3838cc546cca6028d3b73ffcac0d52
`

neheb:
Ah yeah should have updated it. The goal was actually to get rid of all uses of PKG_MD5SUM in the tree/feeds.

PR #529, opened 2019-11-11
batman-adv and friends: improve version string

adrianschmutzler:
This changes the package version string so it does not start
with "openwrt", but with the base version we are modifying:

So far: openwrt-2019.4-1
Now:    2019.4-openwrt-1

Since it's us modifying version 2019.4 (in this case), this order
is more convenient (and also closer to what the kernel version
string does).

### DIFF
--- a/alfred/Makefile
+++ b/alfred/Makefile
@@ -59,7 +59,7 @@ MAKE_FLAGS += \
 	CONFIG_ALFRED_CAPABILITIES=n \
         LIBNL_NAME="libnl-tiny" \
         LIBNL_GENL_NAME="libnl-tiny" \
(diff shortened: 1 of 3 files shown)


adrianschmutzler:
@ecsv @simonwunderlich

adrianschmutzler:
@ecsv @simonwunderlich Ping.

adrianschmutzler:
One could also think about a more kernel-like 2019.4+openwrt1

ecsv:
> One could also think about a more kernel-like 2019.4+openwrt1

This looks somewhat good. But I am not the person merging these things.

The current scheme is more inspired how Debian marks there kernel versions (with "Debian " in front of the version number):

    $ uname -v
    #1 SMP Debian 4.19.67-2+deb10u2 (2019-11-11)

PR #543, opened 2020-01-11
ahcpd: use SHA256 hash and use HTTPS everywhere

BKPepe:
- The old page redirects to a new one and it uses HTTPS.
Let's skip that redirect in PKG_SOURCE_URL and URL.
- Reorder some things to be sync with Makefiles in packages feed
- Add PKG_LICENSE_FILES

Fixes: 47edf2d27594083f91777913bbe57c43a2927f09 (ahcpd: Replace
PKG_MD5SUM with PKG_HASH)
Signed-off-by: Josef Schlehofer <pepe.schlehofer@gmail.com>

### DIFF
--- a/ahcpd/Makefile
+++ b/ahcpd/Makefile
@@ -9,13 +9,15 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ahcpd
 PKG_VERSION:=0.53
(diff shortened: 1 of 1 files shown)


BKPepe:
Compile tested: Turris Omnia, mvebu, OpenWrt master.

PR #546, opened 2020-01-28
bird2: fix service restart and reload

vfreex:
`/etc/init.d/bird restart` or `/etc/init.d/bird reload` has no effects.

This PR fixes this issue by:

- running the service in the foreground to meet the requirements of procd
- sending SIGHUP signal to reload the service

### DIFF
--- a/bird2/Makefile
+++ b/bird2/Makefile
@@ -8,7 +8,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=bird2
 PKG_VERSION:=2.0.7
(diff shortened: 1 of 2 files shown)


vfreex:
Hi, we are running bird2 on OpenWrt. Without proper service reload/restart, we have to do that manually.  Can anyone review this change? @zorun @simonwunderlich @mwarning @axn

zorun:
The maintainer is @tohojo

vfreex:
@tohojo Version bumped. Thank you!

PR #553, opened 2020-03-06
batman-adv 2020.0

ecsv:
@simonwunderlich

----

batman-adv 2020.0
=================

* support latest kernels (3.16 - 5.6)
* use wifi tx rates as fallback for the B.A.T.M.A.N. V throughput estimation
* disable deprecated sysfs support by default
* drop compat code for Linux < 4.14
* replace SUBDIRS with M in kernel recipe
* bugs squashed:
  * fix crash during the scheduling of OGMs for removed interfaces

alfred 2020.0
=============

* fix build against gpsd API 9.0


adrianschmutzler:
If you adjust the package templates here anyway, consider just moving the 3 packages here to openwrt/packages by opening a Pull Request there afterwards.

alfred/Makefile
```diff
@@ -8,9 +8,9 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=alfred
-PKG_VERSION:=2019.5
+PKG_VERSION:=2020.0
 PKG_RELEASE:=0
```
adrianschmutzler: Strictly, PKG_RELEASE starts at "1", not at "0". Since we are changing PKG_VERSION here, which would normally trigger PKG_RELEASE _reset_ to "1", I'd be happy if you could update PKG_RELEASE to 1 as well (for all three packages).

If you do so, please also add one (trivial) line of explanation to the commit messages ...

ecsv: The PKG_RELEASE number change is in another commit. The commit message contains the changes for this release

batctl/Makefile
```diff
@@ -1,54 +1,47 @@
-#
-# Copyright (C) 2014 OpenWrt.org
-#
-# This is free software, licensed under the GNU General Public License v2.
-# See /LICENSE for more information.
-#
+# SPDX-License-Identifier: GPL-2.0
```
adrianschmutzler: Do we need this if PKG_LICENSE is defined below anyway?

ecsv: I understand the PKG_LICENSE as too much reduced debian/copyright. But this information here is about the Makefile

batctl/Makefile
```diff
@@ -1,54 +1,47 @@
...
 
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=batctl
-
 PKG_VERSION:=2020.0
-PKG_RELEASE:=0
-PKG_HASH:=60efe9b148f66aa1b29110493244dc9f1f1d722e6d96969e4d4b2c0ab9278104
+PKG_RELEASE:=1
```
adrianschmutzler: I do not think the changes here require a bump of PKG_RELEASE, as they won't affect the content of the package if I read correctly.

ecsv: You misunderstand again. The PKG_RELEASE here is to follow the PKG_RELEASE numbering of the OpenWrt package feed.

batctl/Makefile
```diff
@@ -1,54 +1,47 @@
...
+  It is an easier method for configuring batman-adv and provides some
+  additional tools for debugging as well. This package builds
+  version $(PKG_VERSION) of the user space utility.
 endef
 
 define Package/batctl-tiny
 $(call Package/batctl/Default)
   TITLE:=B.A.T.M.A.N. Advanced user space configuration tool (Minimal)
   VARIANT:=tiny
-  PROVIDES:=batctl
```
adrianschmutzler: This is not related to the subject of the commit.
So, strictly, this belongs into a separate commit.

However, it could be enough to just add a "While at it, remove redundant PROVIDES for batctl-tiny." to the commit message of this commit depending on how strict rules are applied in this repo.

batctl/Makefile
adrianschmutzler: Okay, I was just asking here, do as you prefer.

batctl/Makefile
adrianschmutzler: Ah, okay. However, in this case I'd still prefer to have the PKG_RELEASE bumped already with the PKG_VERSION increase.

Nevertheless, note that I have no power here, so I'm just suggesting what I think would be best, while the last word is left to the package maintainers. :-)

adrianschmutzler:
Thanks for applying the requested changes, and sorry for not responding on that again, but GitHub didn't notify me of the push.

PR #556, opened 2020-03-06
lede-17.01: batman-adv: Merge bugfixes from 2020.0

ecsv:
@simonwunderlich

----

batman-adv
==========

* Don't schedule OGM for disabled interface


adrianschmutzler:
17.01 is EOL.
I won't stop anyone from merging this, but I do not really see a benefit it. Despite, adding stuff to packages might give the user the impression that 17.01 is still maintained, while the main repo does not receive any (even security) updates anymore.

PR #561, opened 2020-04-22
babeld: Update to version 1.9.2

fblaese:
21 April 2020: babeld-1.9.2

  * Fixed two issues that could cause IPv4 routes to be represented
    incorrectly, with a range of confusing symptoms.  Thanks to
    Fabian Bläse.
  * Fixed incorrect parsing of TLVs with an unknown Address Encoding.
    Thanks to Théophile Bastian.
  * Fixed access to mis-aligned data structure.  Thanks to Antonin Décimo.

Signed-off-by: Fabian Bläse <fabian@blaese.de>

### DIFF
--- a/babeld/Makefile
+++ b/babeld/Makefile
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=babeld
-PKG_VERSION:=1.9.1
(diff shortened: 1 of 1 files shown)


zorun:
Merged, thanks!

PR #571, opened 2020-05-19
batman-adv: use KERNEL_MAKE_FLAGS for kernel module compilation

ecsv:
It is easier to use the global define than to manually keep track of the changes in the various kernel module makefiles in the main OpenWrt repository and feeds. One effect should be that the absolute build_dir information is no longer included in the batman-adv.ko

Also the mac80211's autoconf.h is included to avoid ABI incompatibilities between mac80211 and batman-adv.


adrianschmutzler:
I wonder whether the second one should be considered for backporting?

aparcar:
This makes the module reproducible, big fan.

Issue #547, opened 2020-02-11
gpsd changes result in olsrd being unbuildable


RussellSenior:
gpsd interface changes in the way timestamping works has broken the position update distribution plugin in olsrd. The workaround (I don't need pud) is to remove pud, but it really ought to be patched up.  Here's my horrible hack:

````
index 5f4e1e3..ce82027 100644
--- a/olsrd/Makefile
+++ b/olsrd/Makefile
@@ -116,12 +116,6 @@ define Package/olsrd-mod-pgraph
   TITLE:=output network topology for pgraph
 endef
 
-define Package/olsrd-mod-pud
-  $(call Package/olsrd/template)
-  DEPENDS:=olsrd +libgps
-  TITLE:=Position Update Distribution plugin
-endef
-
 define Package/olsrd-mod-quagga
   $(call Package/olsrd/template)
   DEPENDS:=olsrd
@@ -152,10 +146,6 @@ define Package/olsrd-mod-watchdog
   TITLE:=Watchdog plugin
 endef
 
-define Package/olsrd-mod-pud/conffiles
-/etc/olsrd.d/olsrd.pud.position.conf
-endef
-
 define Package/olsrd-mod-secure/conffiles
 /etc/olsrd.d/olsrd_secure_key
 endef
@@ -170,7 +160,7 @@ MAKE_FLAGS+= \
        DESTDIR="$(PKG_INSTALL_DIR)" \
        STRIP="true" \
        INSTALL_LIB="true" \
-       SUBDIRS="arprefresh bmf dot_draw dyn_gw dyn_gw_plain httpinfo jsoninfo netjson mdns nameservice p2pd pgraph pud quagga secure sgwdynspeed txtinfo watchdog"
+       SUBDIRS="arprefresh bmf dot_draw dyn_gw dyn_gw_plain httpinfo jsoninfo netjson mdns nameservice p2pd pgraph quagga secure sgwdynspeed txtinfo watchdog"
 
 define Build/Compile
        $(call Build/Compile/Default,all)
@@ -252,15 +242,6 @@ define Package/olsrd-mod-pgraph/install
        $(INSTALL_BIN) $(PKG_BUILD_DIR)/lib/pgraph/olsrd_pgraph.so.* $(1)/usr/lib/
 endef
 

````

cotequeiroz:
[...]

PR #575, opened 2020-06-07
olsrd: mark as broken

SvenRoederer:
In October 2018 upstream started looking for a new maintainer and was going into "freeze" mode. No maintainer was found by February 2020 and the code has gone into "hard freeze", which in fact changed it to completely unmaintained.
Since February 2020 there is also issue #547, which is also stalled.

For this reason we mark this package as broken, till a new maintainer is found or it gets dropped completely.

### DIFF
--- a/olsrd/Makefile
+++ b/olsrd/Makefile
@@ -36,7 +36,7 @@ endef
 define Package/olsrd
   $(call Package/olsrd/template)
   MENU:=1
-  DEPENDS:=+libpthread
+  DEPENDS:=+libpthread @BROKEN
 endef
 
 define Package/olsrd/conffiles


SvenRoederer:
For reference the relevant posts on the Olsr Mailinglist:
* https://lists.olsr.org/pipermail/olsr-users/2018-October/007137.html
* https://lists.olsr.org/pipermail/olsr-users/2020-February/007149.html

PolynomialDivision:
Maybe we can just maintain it and accept PRs and stuff like this? Better than throwing away the package completely?

adrianschmutzler:
I don't know how this is handled here, but I'd consider to also drop/update the maintainer for the package then.

SvenRoederer:
@zioproto  as you are listed as maintainer: What do you think?

booo:
I don't think we should mark the whole package as broken. Only the build process of one of the plugins is broken. Just building olsrd works like a charm!

PolynomialDivision:
Package is no longer broken
https://github.com/openwrt-routing/packages/pull/580

zorun:
Looks like olsrd got new maintainers: https://github.com/OLSR/olsrd/pull/79

PolynomialDivision:
Please close this issue. :)

PR #580, opened 2020-06-18
olsrd: bump to latest version

PolynomialDivision:
Fixes https://github.com/openwrt-routing/packages/issues/547


zorun:
Looks good!

Since this is bringing several changes (14 upstream commits), it would be good to have more testing before adding this to 19.07.

PR #581, opened 2020-06-19
luci-app-cjdns: Add rpcd acl file

wfleurant:
Cherry pick of 75c8169a9495fc2059e176c576b6a329dc26672c

Resolved conflicts:
- package release bumped to v7 (snapshots package is version v8)
- backport: GPL, Maintainer, rpcd acl fix for LuCI access

### DIFF
--- a/luci-app-cjdns/Makefile
+++ b/luci-app-cjdns/Makefile
@@ -18,9 +18,10 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=luci-app-cjdns
 PKG_VERSION:=1.3
-PKG_RELEASE:=6
+PKG_RELEASE:=7
 
-PKG_LICENSE:=GPL-3.0
+PKG_MAINTAINER:=William Fleurant <meshnet@protonmail.com>
+PKG_LICENSE:=GPL-3.0-or-later
 
 include $(INCLUDE_DIR)/package.mk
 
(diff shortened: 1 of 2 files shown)


wfleurant:
is it OK to do `git cherry-pick 75c8169` <then edit>  (commit was taken from openwrt-routing trunk/snapshots)
  
or should i have just hand patched from branch `openwrt-19.07`

wfleurant:
@jow- is this the proper way to backport a fix to 19.07? Should the patch be e-mailed?

I also noticed (at least in snapshots) the services->cjdns menu section is exposed at the login screen.  
There are other apps exhibiting this same thing but nothing is accessible until the user is authenticated.

Getting this moved to the new model is overdue.

wfleurant:
I don't think this is the only fix necessary to get the app working in 19.07. I'm testing it now and hitting  
> /usr/lib/lua/luci/dispatcher.lua:1329: module 'luci.cbi' not found:

May be letting this stagnate is my motivation to make the port in to openwrt/luci :exploding_head: :hourglass_flowing_sand:

jow-:
@wfleurant - sorry, I only noticed this now while cleaning up my mailbox. So in principle the backport looks fine. The `module 'luci.cbi' not found` can be solved by adding `luci-compat` to `DEPENDS`

wfleurant:
Tested on OpenWrt 21.02 rpi4 and the admin page is functioning correctly.

PR #587, opened 2020-07-21
batctl: Enable build of mcast_flags subcommand

ecsv:
The mcast_flags subcommand allows to query the mcast_flags of the current device and of the seen originators. It should be enabled for the default and full variants. But the configuration string wasn't correctly included in the list when the variants were prepared and thus disabled in all variants.

Reported-by: Linus Lüssing <linus.luessing@c0d3.blue>
Fixes: 129986825219 ("batctl: Provide different variants")

----

@simonwunderlich

### DIFF
--- a/batctl/Makefile
+++ b/batctl/Makefile
@@ -4,7 +4,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=batctl
 PKG_VERSION:=2020.2
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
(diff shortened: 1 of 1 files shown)


batctl/Makefile
```diff
@@ -154,7 +154,7 @@ config-tables := \
 	claimtable \
 	dat_cache \
 	gateways \
-	loglevel \
```
adrianschmutzler, line 157: Is this removed intentionally, because I don't read anything about it in the commit message?

ecsv: Yes, this is the "not correctly included" part the commit message describes. The string in this list was "loglevel" but should have been "mcast_flags". And after this change, it is "mcast_flags".

adrianschmutzler: Okay, "not correctly included" sounded to me like the correct string is put in the wrong place ...
Actually, it's the wrong string put in the right place after all. :-)

Maybe just write "But the wrong string was put in the list when ..."?

PR #588, opened 2020-07-21
openwrt-19.07: batctl: Enable build of mcast_flags subcommand

ecsv:
The mcast_flags subcommand allows to query the mcast_flags of the current device and of the seen originators. It should be enabled for the default and full variants. But the configuration string wasn't correctly included in the list when the variants were prepared and thus disabled in all variants.

Reported-by: Linus Lüssing <linus.luessing@c0d3.blue>
Fixes: 129986825219 ("batctl: Provide different variants")

----

@simonwunderlich


mwarning:
@simonwunderlich

ecsv:
close unmerged

PR #596, opened 2020-08-11
olsrd: fix compilation with musl 1.2.x

neheb:


mwarning:
@zioproto

zioproto:
@neheb This package is using a git URL for the upstream source code. Why not patching directly the upstream ?

zioproto:
OK I see that the upstream is:
OLSR.org main repository - olsrd v1 - v1 IS IN HARD FREEZE AND NO LONGER MAINTAINED

PR #605, opened 2020-08-27
olsrd: migrate to using bison 3.7.1

pmelange:
With the upgrade to bison 3.7.1 (openwrt commit 1cf842d) building
olsrd was failing.  Now, instead of the contents of header files
being directly inserted into the generated source files, they are
instead included with a #include directive.

The local.mk has, until now, done some magic with *-tmp files,
which is not longer necessary and even causes builds to fail.

src/cfgparser/oparse.c:265:10: fatal error: oparse.h-tmp: No such file or directory
 #include "oparse.h-tmp"

Suggested-by: Jo-Philipp Wich <jo@mein.io>
Signed-off-by: Perry Melange <isprotejesvalkata@gmail.com>

### DIFF
--- a/olsrd/Makefile
+++ b/olsrd/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=olsrd
 PKG_SOURCE_DATE:=2020-06-18
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_SOURCE_PROTO:=git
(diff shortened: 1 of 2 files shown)


PolynomialDivision:
I think the parched should be numbered 100 since we wait for upstream merge?

pmelange:
Well, the musl changes (patch 010) has a PR upstream https://github.com/OLSR/olsrd/pull/84.  This change has no PR upstream .  So, I guess 011 is fine.

PR #606, opened 2020-08-27
olsrd:  fix compilation with bison version 3.7.1

PolynomialDivision:
Recent bison versions changed the way the parser output is generated, which causes compilation of the generated oparse.c to fail with

```
[CC] src/cfgparser/oparse.c
src/cfgparser/oparse.c:265:10: fatal error: oparse.h-tmp: No such file
or directory
 #include "oparse.h-tmp"
          ^~~~~~~~~~~~~~
```

Fix the issue by generating the expected output files directly and by modifying the *.c file in-place.

Suggested-by: Jo-Philipp Wich <jo@mein.io>

### DIFF
--- a/olsrd/Makefile
+++ b/olsrd/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=olsrd
 PKG_SOURCE_DATE:=2020-06-18
-PKG_RELEASE:=2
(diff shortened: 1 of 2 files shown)


pmelange:
Great minds think alike https://github.com/openwrt-routing/packages/pull/605

PolynomialDivision:
Lol xD I will close.

PR #608, opened 2020-09-03
oonf: update olsrv2 packages

PolynomialDivision:


PolynomialDivision:
I tested the v15.01, today (just routing and lan_import). Works nice! Not sure, why this was not updated to the newest stable?

PolynomialDivision:
Ping? Noone wants to merge?

neheb:
https://downloads.openwrt.org/snapshots/faillogs/mips64_octeonplus/routing/oonf-olsrd2/compile.txt

PolynomialDivision:
> https://downloads.openwrt.org/snapshots/faillogs/mips64_octeonplus/routing/oonf-olsrd2/compile.txt

Whups. I think it has something to do with the `010-gcc10.patch`. If I remove it, olsrd2 compiles.
Not sure why my buildsystem just skipped the patch when I compiled it. :/

I will have to test on gcc10 again.

PolynomialDivision:
@neheb Do u think adding `-fcommon` is fine to stay compatible with gcc10?

PR #611, opened 2020-10-05
oonf-olsrd2: fix compiling

PolynomialDivision:
Remove 010-gcc10.patch.
Add `-fcommon` to stay compatible with gcc10. This should only be a temporary fix since it could introduce speed and code size penalty on global variable references.


PolynomialDivision:
@neheb

neheb:
I have a better solution. I'll post it sometime.

neheb:
https://github.com/openwrt-routing/packages/pull/612

PR #612, opened 2020-10-05
oonf-olsrd2: fix static problems

neheb:
The GCC 10 patch marked several variables in headers as extern, which
is correct. However some upstream change marked the definitions as
static even though the definition was not local to the c files.


PolynomialDivision:
Thanks! :)

PolynomialDivision:
can confirm that it builds again with standard openwrt toolchain gcc version.

PR #616, opened 2020-10-06
quagga: update to 1.2.4

neheb:
Refreshed patches. Removed one since OpenWrt no longer uses union
mounts. Fixed up gcc10 patch as well.


mwarning:
thanks for the cleanup!

PR #620, opened 2020-10-24
openwrt-19.07: batman-adv: Merge bugfixes from 2020.4

ecsv:
@simonwunderlich

---

batman-adv
==========

* bla: fix type misuse for backbone_gw hash indexing
* mcast/TT: fix wrongly dropped or rerouted packets
* Add missing include for in_interrupt()
* mcast: fix duplicate mcast packets in BLA backbone from LAN
* mcast: fix duplicate mcast packets in BLA backbone from mesh
* mcast: fix duplicate mcast packets from BLA backbone to mesh

batctl
======

* tcpdump: Fix endianness in ICMPv6 Echo Request/Reply parsing


adschm:
Cool, this might finally resolve my "sent from same source address" warnings ...

adschm:
> Cool, this might finally resolve my "sent from same source address" warnings ...

Unfortunately, they seem to be a different problem, as the fix does not help here.

T-X:
@adschm yeah, some of these when using BLA could cause such an issue. The other issue with "sent from same source address" I'm aware of that at least in my KVM instances somehow bridge interfaces with the same name always gets the same MAC address in different instances. Which is weird, they should be random. But didn't dig deeper into this yet.

If you have a totally different issue, feel free to open a bug report on https://open-mesh.org/ with a description of your setup.

PR #628, opened 2020-12-16
hnetd: update to latest commit

neheb:
Fixed compilation with newer libubox.

libubox added a sys/stat.h include, which happens to break compilation.
The reason being __unused is a variable in musl and an attribute here.
Fixed by undefining right before including the headers.

Clean up Makefile for consistency between packages.

Refreshed patches.


neheb:
ping @mwarning 

Fixes: https://downloads.openwrt.org/snapshots/faillogs/aarch64_cortex-a53/routing/hnetd/nossl/compile.txt

PR #630, opened 2020-12-17
babeld: add ubus bindings

PolynomialDivision:
A better integration of babeld with OpenWrt is to connect the daemon to the ipc. So far, we can only communicate via a websocket. With ubus we can send and receive commands in json format.

### DIFF
--- a/babeld/patches/600-add-ubus.patch
+++ b/babeld/patches/600-add-ubus.patch
@@ -1,0 +1,93 @@
+--- a/babeld.c
++++ b/babeld.c
+@@ -55,6 +55,8 @@ THE SOFTWARE.
+ #include "rule.h"
+ #include "version.h"
+ 
++#include "ubus.h"
++
+ struct timeval now;
+ 
+ unsigned char myid[8];
+@@ -536,6 +538,9 @@ main(int argc, char **argv)
+         }
+     }
+ 
++    if(ubus_bindings)
++        babeld_add_ubus();
++
+     init_signals();
+     rc = resize_receive_buffer(1500);
+     if(rc < 0)
+@@ -635,6 +640,8 @@ main(int argc, char **argv)
+                 FD_SET(local_sockets[i].fd, &readfds);
+                 maxfd = MAX(maxfd, local_sockets[i].fd);
+             }
++            if(ubus_bindings)
++                maxfd = babeld_ubus_add_read_sock(&readfds, maxfd);
+             rc = select(maxfd + 1, &readfds, NULL, NULL, &tv);
+             if(rc < 0) {
+                 if(errno != EINTR) {
+@@ -703,6 +710,9 @@ main(int argc, char **argv)
+             i++;
+         }
+ 
++        if(ubus_bindings)
++            babeld_ubus_receive(&readfds);
++
+         if(reopening) {
+             kernel_dump_time = now.tv_sec;
+             check_neighbours_timeout = now;
+--- a/Makefile
++++ b/Makefile
+@@ -11,11 +11,11 @@ LDLIBS = -lrt
+ 
+ SRCS = babeld.c net.c kernel.c util.c interface.c source.c neighbour.c \
+        route.c xroute.c message.c resend.c configuration.c local.c \
+-       disambiguation.c rule.c
++       disambiguation.c rule.c ubus.c
+ 
+ OBJS = babeld.o net.o kernel.o util.o interface.o source.o neighbour.o \
+        route.o xroute.o message.o resend.o configuration.o local.o \
+-       disambiguation.o rule.o
++       disambiguation.o rule.o ubus.o
+ 
+ babeld: $(OBJS)
+ 	$(CC) $(CFLAGS) $(LDFLAGS) -o babeld $(OBJS) $(LDLIBS)
+--- a/generate-version.sh
++++ b/generate-version.sh
+@@ -10,4 +10,4 @@ else
+     version="unknown"
+ fi
+ 
+-echo "#define BABELD_VERSION \"$version\""
++echo "#define BABELD_VERSION \"$version-ubus-mod\""
+--- a/configuration.c
++++ b/configuration.c
+@@ -41,6 +41,7 @@ THE SOFTWARE.
+ #include "kernel.h"
+ #include "configuration.h"
+ #include "rule.h"
++#include "ubus.h"
+ 
+ static struct filter *input_filters = NULL;
+ static struct filter *output_filters = NULL;
+@@ -850,7 +851,8 @@ parse_option(int c, gnc_t gnc, void *clo
+               strcmp(token, "daemonise") == 0 ||
+               strcmp(token, "skip-kernel-setup") == 0 ||
+               strcmp(token, "ipv6-subtrees") == 0 ||
+-              strcmp(token, "reflect-kernel-metric") == 0) {
++              strcmp(token, "reflect-kernel-metric") == 0 ||
++              strcmp(token, "ubus-bindings") == 0) {
+         int b;
+         c = getbool(c, &b, gnc, closure);
+         if(c < -1)
+@@ -868,6 +870,8 @@ parse_option(int c, gnc_t gnc, void *clo
+             has_ipv6_subtrees = b;
+         else if(strcmp(token, "reflect-kernel-metric") == 0)
+             reflect_kernel_metric = b;
++        else if(strcmp(token, "ubus-bindings") == 0)
++            ubus_bindings = b;
+         else
+             abort();
+     } else if(strcmp(token, "protocol-group") == 0) {
--- a/babeld/src/ubus.c
+++ b/babeld/src/ubus.c
@@ -1,0 +1,384 @@
+#include <stdint.h>
+#include <stdlib.h>
+#include <sys/select.h>
+
+#include <libubox/blob.h>
+#include <libubox/blobmsg.h>
+#include <libubox/list.h>
+
+#include <arpa/inet.h>
+#include <net/if.h>
+#include <netinet/in.h>
+#include <sys/ioctl.h>
+#include <sys/socket.h>
+
+#include "babeld.h"
+#include "configuration.h"
+#include "interface.h"
+#include "kernel.h"
+#include "local.h"
+#include "message.h"
+#include "neighbour.h"
+#include "net.h"
+#include "resend.h"
+#include "route.h"
+#include "rule.h"
+#include "source.h"
+#include "util.h"
+#include "version.h"
+#include "xroute.h"
+
+#include "ubus.h"
+
+// Definition of header variable whether to enable ubus bindings.
+int ubus_bindings = 0;
+
+// Shared state maintained throughout calls to handle ubus messages.
+static struct ubus_context *shared_ctx;
+
+// List of exported routes (to be used with ubox's list helpers).
+struct xroute_list_entry {
+  struct list_head list;
+  struct xroute *xroute;
+};
+
+// List of received routes (to be used with ubox's list helpers).
+struct route_list_entry {
+  struct list_head list;
+  struct route_stream *route;
+};
+
+// List of neighbours (to be used with ubox's list helpers).
+struct neighbour_list_entry {
+  struct list_head list;
+  struct neighbour *neighbour;
+};
+
+// Sends a babel info message on ubus socket.
+static void babeld_ubus_babeld_info(struct ubus_context *ctx_local,
+                                    struct ubus_object *obj,
+                                    struct ubus_request_data *req,
+                                    const char *method, struct blob_attr *msg) {
+  struct blob_buf b = {0};
+  void *prefix;
+  char host[64];
+  int ret;
+
+  blob_buf_init(&b, 0);
+  blobmsg_add_string(&b, "babeld-version", BABELD_VERSION);
+  blobmsg_add_string(&b, "my-id", format_eui64(myid));
+  if (!gethostname(host, sizeof(host)))
+    blobmsg_add_string(&b, "host", host);
+
+  ret = ubus_send_reply(ctx_local, req, b.head);
+  if (ret)
+    fprintf(stderr, "Failed to send reply: %s\n", ubus_strerror(ret));
+}
+
+// Appends an exported route message entry to the buffer.
+static void babeld_add_xroute_buf(struct xroute *xroute, struct blob_buf *b) {
+  void *prefix;
+
+  prefix = blobmsg_open_table(b, format_prefix(xroute->prefix, xroute->plen));
+
+  blobmsg_add_string(b, "src-prefix",
+                     format_prefix(xroute->src_prefix, xroute->src_plen));
+  blobmsg_add_u32(b, "metric", xroute->metric);
+  blobmsg_close_table(b, prefix);
+}
+
+// Sends an exported routes message on ubus socket, splitting apart IPv4 and IPv6 routes.
+static void babeld_ubus_get_xroutes(struct ubus_context *ctx_local,
+                                    struct ubus_object *obj,
+                                    struct ubus_request_data *req,
+                                    const char *method, struct blob_attr *msg) {
+  struct blob_buf b = {0};
+  struct xroute_stream *xroutes;
+  struct xroute_list_entry *cur, *tmp;
+  void *ipv4, *ipv6;
+  int ret;
+  LIST_HEAD(xroute_ipv4_list);
+  LIST_HEAD(xroute_ipv6_list);
+
+  xroutes = xroute_stream();
+  if (xroutes) {
+    while (1) {
+      struct xroute *xroute = xroute_stream_next(xroutes);
+      if (xroute == NULL)
+        break;
+
+      struct xroute_list_entry *xr =
+          calloc(1, sizeof(struct xroute_list_entry));
+      xr->xroute = xroute;
+
+      if (v4mapped(xroute->prefix)) {
+        list_add(&xr->list, &xroute_ipv4_list);
+      } else {
+        list_add(&xr->list, &xroute_ipv6_list);
+      }
+    }
+    xroute_stream_done(xroutes);
+  }
+
+  blob_buf_init(&b, 0);
+  ipv4 = blobmsg_open_table(&b, "IPv4");
+  list_for_each_entry_safe(cur, tmp, &xroute_ipv4_list, list) {
+    babeld_add_xroute_buf(cur->xroute, &b);
+    list_del(&cur->list);
+    free(cur);
+  }
+  blobmsg_close_table(&b, ipv4);
+
+  ipv6 = blobmsg_open_table(&b, "IPv6");
+  list_for_each_entry_safe(cur, tmp, &xroute_ipv6_list, list) {
+    babeld_add_xroute_buf(cur->xroute, &b);
+    list_del(&cur->list);
+    free(cur);
+  }
+  blobmsg_close_table(&b, ipv6);
+
+  ret = ubus_send_reply(ctx_local, req, b.head);
+  if (ret)
+    fprintf(stderr, "Failed to send reply: %s\n", ubus_strerror(ret));
+}
+
+// Appends an route message entry to the buffer.
+static void babeld_add_route_buf(struct babel_route *route,
+                                 struct blob_buf *b) {
+  void *prefix;
+  char channels[100];
+
+  if (route->channels_len == 0) {
+    channels[0] = '\0';
+  } else {
+    int i, j = 0;
+    snprintf(channels, sizeof(channels), " chan (");
+    j = strlen(channels);
+    for (i = 0; i < route->channels_len; i++) {
+      if (i > 0)
+        channels[j++] = ',';
+      snprintf(channels + j, sizeof(channels) - j, "%u",
+               (unsigned)route->channels[i]);
+      j = strlen(channels);
+    }
+    snprintf(channels + j, sizeof(channels) - j, ")");
+  }
+
+  prefix = blobmsg_open_table(
+      b, format_prefix(route->src->prefix, route->src->plen));
+
+  blobmsg_add_string(
+      b, "src-prefix",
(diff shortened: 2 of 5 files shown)


babeld/src/ubus.h
```diff
@@ -0,0 +1,10 @@
+
+#include <libubus.h>
+
+void babeld_add_ubus();
+
+int babeld_ubus_add_read_sock(fd_set* readfds, int maxfd);
+
+int babeld_ubus_isset_ubus(fd_set* readfds);
+
+void babeld_ubus_receive();
```
adschm, line 10: missing newline at EOF ...

babeld/src/ubus.h
```diff
@@ -0,0 +1,10 @@
+
```
adschm, line 1: this empty line can be dropped?

PolynomialDivision:
@adschm Thanks a lot for all the comments already. :) I added now the first part of the "dump" to show "my-id". But acutally, my id is always changing?

```
root@OpenWrt:/tmp# ubus call babeld dump
{
	"my-id": 4334176
}
root@OpenWrt:/tmp# ubus call babeld dump
{
	"my-id": 4334204
}
root@OpenWrt:/tmp# ubus call babeld dump
{
	"my-id": 4334232
}
root@OpenWrt:/tmp# ubus call babeld dump
{
	"my-id": 4334260
}
root@OpenWrt:/tmp# ubus call babeld dump
{
	"my-id": 4334176
}
root@OpenWrt:/tmp# ubus call babeld dump
{
	"my-id": 4334204
}
```
Any idea? I will have a deeper look into this.

PolynomialDivision:
Lol, the id is a string. I will fix everxthing and mention you again, then I'm finished.

PolynomialDivision:
The only thing that I'm still not sure is that I would like to check folr `v4mapped(prefix)` and divide into v4 and v6 addresses. But not sure how to do that.

PolynomialDivision:
I'm adding that list stuff.

kerneis:
This seems like a significant addition rather than a mere patch. I wonder if it would make sense to try and upstream it, or at least ask @jech how he feels about it. Also, ubus.h and ubus.cc should really be part of the patch file rather than added directly to the sources in my opinion.

PolynomialDivision:
@kerneis It is a very common procedure to add ubus bindings to OpenWrt in that way.
A famous example ist hostapd:
https://github.com/openwrt/openwrt/tree/master/package/network/services/hostapd



> Also, ubus.h and ubus.cc should really be part of the patch file rather than added directly to the sources in my opinion.

I would point to some drivers, or again the hostapd. I think having such a big patchfile is not that easy. And maintainance is better in seperated files.

>  I wonder if it would make sense to try and upstream it,

I would keep it in OpenWrt, since no other OS has ubus bindings. But if @jech want, I can add this also upstream. :)

I hope that noone is against adding ubus bindings. I added only basic functionality and want to add more stuff. :) I think this will enable great features and new opportunities. (Soon, I will publish more code)

PolynomialDivision:
@adschm I think I'm finished. I'm sorting now ipv4 and ipv6.
I have now the calls:

```
root@OpenWrt:~# ubus call babeld get_info
{
	"babeld-version": "babeld-1.9.2",
	"my-id": "32:xx:xx:xx:xx:xx:a3:7d",
	"host": "OpenWrt"
}
```

```
root@OpenWrt:~# ubus call babeld get_routes
{
	"IPv4": {
		"10.2.0.1/32": {
			"src-prefix": "0.0.0.0/0",
			"route_metric": 96,
			"route_smoothed_metric": 96,
			"refmetric": 0,
			"id": "62:e3:xx:xx:xx:fb:4e:a6",
			"seqno": 41381,
			"channels": "",
			"age": 17,
			"via": "fe80::xx:xxff:fxx:4ea6",
			"nexthop": " nexthop ",
			"installed": true,
			"feasible": true
		}
	"IPv6": {
		
	}
}

```
root@OpenWrt:~# ubus call babeld get_xroutes
{
	"IPv4": {
		"10.0.0.3/32": {
			"src-prefix": "0.0.0.0/0",
			"metric": 0
		},
		"10.0.0.0/24": {
			"src-prefix": "0.0.0.0/0",
			"metric": 0
		}
	},
	"IPv6": {
		"fdfa:xx:xx::1/128": {
			"src-prefix": "::/0",
			"metric": 0
		}
	}
}

```

**Comment by PolynomialDivision** [comment]
Argh, I have to rename all that neighbor stuff... ^^

**Comment by PolynomialDivision** [comment]
@kerneis Are u okay, if I create a separate package, e.g. ubabeld? Then I can maintain the stuff there and go further with it. And at the end we can find a way to merge again?

**Comment by jech** [comment]
Thanks for the work.

> This seems like a significant addition rather than a mere patch. I wonder if it
> would make sense to try and upstream it, or at least ask @jech how he feels
> about it.

I'd rather not accept it upstream, since then I would have to maintain it,
and I no longer have any devices able to run recent releases of OpenWRT
(all my OpenWRT devices have 32MB RAM).

I agree with Gabriel that this is a fairly significant addition.  However,
I find it clean and modular, and wouldn't object to it getting into
OpenWRT.  Please add a description of your changes somewhere, so that
people aren't confused about the divergence.

I have no objection to the fork being called babeld.  I have no objection
to the fork being called ubabeld.  On the other hand, I would like to
avoid the situation in which there are both babeld and ubabeld in OpenWRT,
as that would lead to confusion.

I'm happy to trust Gabriel with making the right decision.

**Comment by PolynomialDivision** [comment]
> Thanks for the work.

> However, I find it clean and modular, and wouldn't object to it getting into OpenWRT.

Thanks it is always nice to hear that people are thankful for your contribution! :D Thanks for making babel. :)

I will rewrite my commit message to add a better description.

**Comment by PolynomialDivision** [comment]
@kerneis anything new? ;)

**Comment by kerneis** [comment]
Based on what @jech said I think it's fine to merge keeping the name babeld. Ideally, the version information (`babeld -V`) would be updated to mention the patch, so that it's easier to know what fork is running when asking users for debug information.

I'd like to review the patch more carefully before it's merged, but won't have time before next week. In the meantime, if somebody can test (I don't have access to OpenWrt devices these days, which probably means I shouldn't maintain this package anymore), I'd be grateful for feedback. @zorun maybe?

Gabriel

**Comment by PolynomialDivision** [comment]
> the version information (`babeld -V`) would be updated to mention the patch, so that it's easier to know what fork is running when asking users for debug information.

Okay. I think I will just add something like
https://github.com/jech/babeld/blob/master/generate-version.sh#L13
`echo "#define BABELD_VERSION \"$version\"-ubus-mod"`

**Comment by PolynomialDivision** [comment]
```
root@OpenWrt:/tmp# ubus call babeld get_info
{
	"babeld-version": "babeld-1.9.2-ubus-mod",
	"my-id": "32:b5:c2:ff:fe:6e:a3:7d",
	"host": "OpenWrt"
}
```

**Comment by zoobab** [comment]
What does ""channel": -2," means in the example?

**Comment by PolynomialDivision** [comment]
> What does ""channel": -2," means in the example?

I think that is a babel magic number for weird conection, but I have no idea. Just guessing.

**Comment by jech** [comment]
Yes.  https://github.com/jech/babeld/blob/master/interface.h#L85

**Comment by PolynomialDivision** [comment]
Something new?
I really would like to go on. :P

File: babeld/src/ubus.c [hunk 3]
```diff
@@ -0,0 +1,357 @@
...
+    UBUS_OBJECT_TYPE("babeld", babeld_methods);
+
+static struct ubus_object babeld_object = {
+    .name = "babeld",
+    .type = &babeld_object_type,
+    .methods = babeld_methods,
+    .n_methods = ARRAY_SIZE(babeld_methods),
+};
+
+static void ubus_add_oject() {
```
kerneis, line 316: typo: ubus_add_object

babeld/src/ubus.c
```diff
@@ -0,0 +1,357 @@
...
+}
+
+int babeld_ubus_isset_ubus(fd_set *readfds) {
+  if (!ctx)
+    return 0;
+  return FD_ISSET(ctx->sock.fd, readfds);
+}
+
+int babeld_ubus_add_read_sock(fd_set *readfds, int maxfd) {
+  FD_SET(ctx->sock.fd, readfds);
```
kerneis, line 350: Check that ctx is not null.

babeld/src/ubus.c
```diff
@@ -0,0 +1,357 @@
...
+    .name = "babeld",
+    .type = &babeld_object_type,
+    .methods = babeld_methods,
+    .n_methods = ARRAY_SIZE(babeld_methods),
+};
+
+static void ubus_add_oject() {
+  int ret;
+
+  ret = ubus_add_object(ctx, &babeld_object);
```
kerneis, line 319: Can you please include the relevant header, I couldn't find it?

kerneis (changes requested):

babeld/src/ubus.c
```diff
@@ -0,0 +1,357 @@
...
+#include "route.h"
+#include "rule.h"
+#include "source.h"
+#include "util.h"
+#include "version.h"
+#include "xroute.h"
+
+#include "ubus.h"
+
+LIST_HEAD(xroute_ipv4_list);
```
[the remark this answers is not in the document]
jech: It was difficult to write, there's no reason it should be easy to read.

babeld/patches/600-add-ubus.patch
```diff
@@ -0,0 +1,63 @@
...
++#include "ubus.h"
++
+ struct timeval now;
+ 
+ unsigned char myid[8];
+@@ -170,6 +172,8 @@ main(int argc, char **argv)
+     change_smoothing_half_life(4);
+     has_ipv6_subtrees = kernel_has_ipv6_subtrees();
+ 
++    babeld_add_ubus();
```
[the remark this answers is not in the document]
PolynomialDivision: You mean an option on babeld start?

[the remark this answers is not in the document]
kerneis: Yes, something similar to the `-g` flag in the regular babeld.

babeld/src/ubus.c
```diff
@@ -0,0 +1,357 @@
...
+  ret = ubus_send_reply(ctx, req, b.head);
+  if (ret)
+    fprintf(stderr, "Failed to send reply: %s\n", ubus_strerror(ret));
+}
+
+static void babeld_add_route_buf(struct babel_route *route) {
+  void *prefix;
+  char channels[100];
+
+  const unsigned char *nexthop =
```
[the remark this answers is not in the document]
PolynomialDivision: that's babel code and not documented :/ not sure why it is needed.

babeld/src/ubus.c
PolynomialDivision: that is #include <libubus.h>

babeld/src/ubus.c
```diff
@@ -0,0 +1,357 @@
...
+
+LIST_HEAD(neighbour_ipv4_list);
+LIST_HEAD(neighbour_ipv6_list);
+
+struct neighbour_list_entry {
+  struct list_head list;
+  struct neighbour *neighbour;
+};
+
+static struct ubus_context *ctx;
```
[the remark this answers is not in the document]
PolynomialDivision: making the `blob_buf b` global is a typical thing in openwrt. But I can make it local.

babeld/patches/600-add-ubus.patch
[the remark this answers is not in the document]
PolynomialDivision: Any idea why this does not work?
https://github.com/openwrt-routing/packages/pull/630/files#diff-d414ab8ec5eed4328cb50c6a16058041e6842cc0821536ced6c61fc3a34673ffR93-R94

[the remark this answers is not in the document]
kerneis: I think you test the value before parsing it in if(ubus_bindings)  babeld_add_ubus().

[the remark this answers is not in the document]
PolynomialDivision: shit. ^^ thanks!!! :)

PolynomialDivision:
@kerneis Thanks for all the good comments! I changed everything.

babeld/src/ubus.c
PolynomialDivision: so it is already included in ubus.h, so I won't double include it in ubus.c?

PolynomialDivision:
rebased on current master

PolynomialDivision:
I added
```
	# option 'ubus_bindings' 'false'
```
to the config.

kerneis:
I will try to give this another round of review quickly (github interface doesn't make it easy, sorry). In the meantime, could anybody else confirm that the latest patch compiles and behaves correctly please? I don't have openwrt devices at home anymore…

babeld/patches/600-add-ubus.patch
```diff
@@ -0,0 +1,97 @@
...
+-echo "#define BABELD_VERSION \"$version\""
++echo "#define BABELD_VERSION \"$version-ubus-mod\""
+--- a/babeld.h
++++ b/babeld.h
+@@ -106,6 +106,9 @@ extern int protocol_socket;
+ extern int kernel_socket;
+ extern int max_request_hopcount;
+ 
++// config option for ubus bindings
++extern int ubus_bindings;
```
kerneis, line 72: Please move this to ubus.h instead: it's easier to think of exported variables in terms of translation units, and to have the declaration of a variable in the header corresponding to the .cc file where it is defined. (Also, this will remove babel.h completely from this patch, which will make it easier to read).

babeld/src/ubus.c
```diff
@@ -0,0 +1,373 @@
...
+#include "route.h"
+#include "rule.h"
+#include "source.h"
+#include "util.h"
+#include "version.h"
+#include "xroute.h"
+
+#include "ubus.h"
+
+// config option for ubus bindings
```
kerneis, line 34: // Whether to enable ubus bindings (boolean option).

babeld/src/ubus.c
```diff
@@ -0,0 +1,373 @@
...
+int ubus_bindings = 0;
+
+// struct definition for sorting xroutes
```
kerneis, line 37: Are we really sorting them? It looked to me we just separate v4 and v6.

// List of exported routes (to be used with ubox's list helpers)

idem below

babeld/src/ubus.c
```diff
@@ -0,0 +1,373 @@
...
+  struct route_stream *route;
+};
+
+// struct definition for sorting neighbours
+struct neighbour_list_entry {
+  struct list_head list;
+  struct neighbour *neighbour;
+};
+
+// pointer to ubus context
```
kerneis, line 55: When commenting, try to focus on the why rather than paraphrasing the type information (eg. here, why is the variable global and what is its lifetime):

// Shared state maintained throughout calls to handle ubus messages.

babeld/src/ubus.c
```diff
@@ -0,0 +1,373 @@
...
+  char host[64];
+  int ret;
+
+  blob_buf_init(&b, 0);
+  blobmsg_add_string(&b, "babeld-version", BABELD_VERSION);
+  blobmsg_add_string(&b, "my-id", format_eui64(myid));
+  if (!gethostname(host, sizeof(host)))
+    blobmsg_add_string(&b, "host", host);
+
+  ret = ubus_send_reply(ctx, req, b.head);
```
kerneis, line 73: I'm confused: do you intend to use ctx or ctx_local here? Maybe they're just the same? Because your initial code used the local one, and since the rename it uses the local one.

Maybe it makes more sense to call the local variables ctx and the global one shared_ctx or persistent_ctx?

babeld/src/ubus.c
```diff
@@ -0,0 +1,373 @@
...
+// struct definition for sorting neighbours
+struct neighbour_list_entry {
+  struct list_head list;
+  struct neighbour *neighbour;
+};
+
+// pointer to ubus context
+static struct ubus_context *ctx;
+
+static void babeld_ubus_babeld_info(struct ubus_context *ctx_local,
```
kerneis, line 58: // Sends a babel info message on ubus socket.

babeld/src/ubus.c
```diff
@@ -0,0 +1,373 @@
...
+  blobmsg_add_string(&b, "my-id", format_eui64(myid));
+  if (!gethostname(host, sizeof(host)))
+    blobmsg_add_string(&b, "host", host);
+
+  ret = ubus_send_reply(ctx, req, b.head);
+  if (ret)
+    fprintf(stderr, "Failed to send reply: %s\n", ubus_strerror(ret));
+}
+
+static void babeld_add_xroute_buf(struct xroute *xroute, struct blob_buf *b) {
```
kerneis, line 78: // Appends an exported route message entry to the buffer.


idem for similar functions below

babeld/src/ubus.c
```diff
@@ -0,0 +1,373 @@
...
+
+  prefix = blobmsg_open_table(b, format_prefix(xroute->prefix, xroute->plen));
+
+  blobmsg_add_string(b, "src-prefix",
+                     format_prefix(xroute->src_prefix, xroute->src_plen));
+  blobmsg_add_u32(b, "metric", xroute->metric);
+  blobmsg_close_table(b, prefix);
+}
+
+static void babeld_ubus_get_xroutes(struct ubus_context *ctx_local,
```
kerneis, line 89: // Sends an exported route message on ubus socket, splitting apart IPv4 and IPv6 routes.

idem for similar functions below

babeld/src/ubus.c
```diff
@@ -0,0 +1,373 @@
...
+    UBUS_OBJECT_TYPE("babeld", babeld_methods);
+
+static struct ubus_object babeld_object = {
+    .name = "babeld",
+    .type = &babeld_object_type,
+    .methods = babeld_methods,
+    .n_methods = ARRAY_SIZE(babeld_methods),
+};
+
+static bool ubus_init_object() {
```
kerneis, line 323: // Registers handlers for babel methods in the global ubus context.

babeld/src/ubus.c
```diff
@@ -0,0 +1,373 @@
...
+  ret = ubus_add_object(ctx, &babeld_object);
+  if (ret) {
+    fprintf(stderr, "Failed to add object: %s\n", ubus_strerror(ret));
+    return false;
+  }
+
+  return true;
+}
+
+static bool babeld_ubus_init(void) {
```
kerneis, line 335: // Initializes the global ubus context, connecting to the bus to be able to receive and send messages.

kerneis (changes requested):
A couple of minor changes still needed, but overall it seems ready for inclusion to me. Thanks for the great work. If nobody else chimes in, I'll merge your PR once the changes are done, and revert if someone complains that it breaks them.

babeld/src/ubus.c
PolynomialDivision: "exported routes", or?

babeld/src/ubus.c
PolynomialDivision: Yes. This was an mistake. But actually it makes no difference as you already explained. :)

PolynomialDivision:
> A couple of minor changes still needed, but overall it seems ready for inclusion to me. Thanks for the great work. If nobody else chimes in, I'll merge your PR once the changes are done, and revert if someone complains that it breaks them.

Thanks for all your efforts. :)

If this is finished, I would like to send the received routes also over the bus. Then I can monitor easily and directly if someone else 
tries to announce certain prefixes. I think I have to do this here https://github.com/jech/babeld/blob/master/route.c#L859
If this patch is merged I can poll for new updates, but with the above mentioned changes I can react to changes more quickly. :)

jech:
What is the performance impact of all of that?

kerneis (approved):

kerneis:
> What is the performance impact of all of that?

I'm not worried too much currently because it's disabled by default, and it has to be polled so even if active the user decides what the cost is. It doesn't look more expensive than the babel local interface (at least asymptotically, it copies quite a lot and I don't know what the overhead is of the ubox function calls).

kerneis:
> If this is finished, I would like to send the received routes also over the bus. Then I can monitor easily and directly if someone else
> tries to announce certain prefixes. I think I have to add some function to the "local_notify..." parts.
> If this patch is merged I can poll for new updates, but with the above mentioned changes I can react to changes more quickly. :)

Two points:
- when working on github pull requests, please create follow-up commits rather than amending, and then squash at the very end of the process. We shouldn't need to do this (gerrit for instance handles the amend pattern very well), but github garbage collects old versions of the patch, so it's impossible to compare changes since the last review if you amend, and it's painful as a reviewer.
- I'd rather have you open a separate bug or pull request to discuss the pushing model. I'm not sure how you intend to do it, because you'll need to push the initial state when a new client starts listening on the bus. Is this something ubox allows?

PolynomialDivision:
> > What is the performance impact of all of that?
> 
> I'm not worried too much currently because it's disabled by default, and it has to be polled so even if active the user decides what the cost is. It doesn't look more expensive than the babel local interface (at least asymptotically, it copies quite a lot and I don't know what the overhead is of the ubox function calls).

This is also my opinion.

For me, it makes things more efficient because I can request only the things I need. The parsing can then also be done very efficiently in c. The only thing I find unnecessary in the code is that when I separate in ipv4 and ipv6 to makes two lists. Sadly, my testbed is not that big right now, but with increasing number of nodes I will look again at performance.

Overall, I think to connect babeld to ubus will make it more attractive to other openwrt-devs, since they are used to use this way of inter process communication. :)

This will not be the last commit you will receive from me. ;) Probably, we will role out the babel mesh daemon over [Berlin Freifunk Network](https://berlin.freifunk.net/), since it is working very smooth in our testbed. :)

> when working on github pull requests, please create follow-up commits rather than amending, and then squash at the very end of the process. We shouldn't need to do this (gerrit for instance handles the amend pattern very well), but github garbage collects old versions of the patch, so it's impossible to compare changes since the last review if you amend, and it's painful as a reviewer.

Okay. :)

> push the initial state when a new client starts listening on the bus.

I have to look into the code to see what u mean with that. I will write some code over the weekend and make a new PR. ;)

PR #633, opened 2021-01-15
babeld: send events via ubus

PolynomialDivision:
Send a notification via the ubus bus if we experience any changes in neighbors, routes or xroutes.

The format looks like this:
 - {route,xroute,neighbour}.add: Object was added
 - {route,xroute,neighbour}.change: Object was changed
 - {route,xroute,neighbour}.flush: Object was flushed

If ubus_bindings is turned off, it will minimally effect performance, since only an if-statement has to be evaluated. If no subscriber is available, it will minimally change the performance, since only an if-statmenet that checks for subscribers has to be evaluated.

### DIFF
--- a/babeld/src/ubus.c
+++ b/babeld/src/ubus.c
@@ -379,7 +379,7 @@ void ubus_notify_route(struct babel_route *route, int kind) {
 
 void ubus_notify_xroute(struct xroute *xroute, int kind) {
   struct blob_buf b = {0};
-  char method[50]; // possible methods are route.change, route.add, route.flush
+  char method[50]; // possible methods are xroute.change, xroute.add, xroute.flush
 
   if (!babeld_object.has_subscribers)
     return;
@@ -398,7 +398,7 @@ void ubus_notify_xroute(struct xroute *xroute, int kind) {
 
 void ubus_notify_neighbour(struct neighbour *neigh, int kind) {
   struct blob_buf b = {0};
-  char method[50]; // possible methods are route.change, route.add, route.flush
+  char method[50]; // possible methods are neigh.change, neigh.add, neigh.flush
 
   if (!babeld_object.has_subscribers)
     return;
--- a/babeld/src/ubus.c
+++ b/babeld/src/ubus.c
@@ -360,7 +360,7 @@ static bool babeld_ubus_init(void) {
 
 void ubus_notify_route(struct babel_route *route, int kind) {
   struct blob_buf b = {0};
-  char method[13]; // max is route.change
+  char method[50]; // possible methods are route.change, route.add, route.flush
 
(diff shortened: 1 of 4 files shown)


PolynomialDivision:
@kerneis This is highly wip, but that is what I mean with notify the ubus bus.

PolynomialDivision:
I try upstreaming the change in babeld:

https://github.com/jech/babeld/pull/72

PolynomialDivision:
@kerneis This is what I mean with sending notifications on the bus. Now I just receive the part I need. Polling would require babeld to create each time probably thousands of messages. With updates I can just check the updates. For example, I can check if a certain prefix is announced, by just looking add the "route.add" methods.

PolynomialDivision:
I wrote some small daemon that shows how to connect to the ubus interface and process the data:
https://github.com/PolynomialDivision/babeld-ids

More logic will follow soon.

PolynomialDivision:
@kerneis

babeld/src/ubus.c
```diff
@@ -354,6 +358,63 @@ static bool babeld_ubus_init(void) {
...
+
+  if (!route)
+    return;
+
+  if (!shared_ctx)
+    return;
+
+  blob_buf_init(&b, 0);
+  babeld_add_route_buf(route, &b);
+  sprintf(method, "route.%s", local_kind(kind));
```
kerneis, line 376: Please use snprintf(method, sizeof(method), ...), I'm paranoid about buffer overflow. Idem below.

babeld/src/ubus.c
```diff
@@ -354,6 +358,63 @@ static bool babeld_ubus_init(void) {
   return true;
 }
 
+void ubus_notify_route(struct babel_route *route, int kind) {
+  struct blob_buf b = {0};
+  char method[13]; // max is route.change
```
kerneis, line 363: char method[sizeof("route.change")]; // possible methods are route.change, route.add, route.flush

Idem below

kerneis: Or just use a largish value and don't bother (50, 100)?

kerneis (approved):
Looks good to me with a few minor comments.

babeld/src/ubus.c
PolynomialDivision: I will just go for 50. ;)

babeld/src/ubus.c
PolynomialDivision: True story. :)

PolynomialDivision:
Probably I will rewrite a lot of stuff again, if this is merged upstream:
https://github.com/jech/babeld/pull/57

kerneis (approved):

PR #638, opened 2021-01-31
alfred: add "network" to reload triggers

adschm:
alfred uses properties from network interfaces, like the MAC
address. Since these might be altered by /etc/config/network,
alfred should be restarted on change.

---

@simonwunderlich @ecsv 

I also think this should be backported, but I'm waiting for the feedback on this PR first.

### DIFF
--- a/alfred/Makefile
+++ b/alfred/Makefile
@@ -4,7 +4,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=alfred
 PKG_VERSION:=2021.0
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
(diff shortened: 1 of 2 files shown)


ecsv:
Could you please point me to a parameter change in /etc/config/network which could cause procd to restart alfred. Because I only see an reload trigger but nothing which would force a restart (like a parameter change in any of the procd_*_instance sections).

Btw. it seems like the script is also broken when it comes to disabling the alfred daemon on `/etc/init.d/alfred reload` (when the "disabled" switches from 0 to 1). In case you want to check this out too. Fix should be to change the check 

    [ "$enable" = 0 ] && exit 1

to

    [ "$enable" = 0 ] && return


But there are various other weird looking exit calls in this init script which should be checked out and modified to NOT simply kill the process (and thus stop the logic for the procd init.d helpers in /lib/functions/procd.sh ).

adschm:
I'm sorry, I actually thought this was a quick shot. Thanks for taking over part of the work.
Unfortunately, I probably won't be able to pick this up for a little longer ...

PR #640, opened 2021-02-01
babeld: adopt to upstream header-style

PolynomialDivision:
Instead of including the headerfiles that define the structs, we add
the struct definitions to our headerfile.

Fixes warning:

```
ubus.h:67:32: warning: 'struct xroute' declared inside parameter list will not be visible outside of this definition or declaration
 void ubus_notify_xroute(struct xroute *xroute, int kind);
                                ^~~~~~
```

### DIFF
--- a/babeld/Makefile
+++ b/babeld/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=babeld
 PKG_VERSION:=1.9.2
-PKG_RELEASE:=5
+PKG_RELEASE:=$(AUTORELEASE)
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://www.irif.fr/~jch/software/files/
--- a/babeld/Makefile
+++ b/babeld/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=babeld
 PKG_VERSION:=1.9.2
-PKG_RELEASE:=4
+PKG_RELEASE:=5
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://www.irif.fr/~jch/software/files/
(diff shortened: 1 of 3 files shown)


kerneis (changes requested):
This seems like the wrong fix to me, but maybe because I come from an institutional culture of C++ where we avoid forward declarations and instead include the headers that we use.

@jech: can you shed some light on the pros/cons of those forward-declarations for structs (is it because otherwise you'd end up with dependency loops)?

PolynomialDivision:
Here is an example of upstream code:
https://github.com/jech/babeld/blob/master/local.h#L23-L25

Since we do not need the size of the structs I think it is fine. I'm also very interested about pros/cons! :)

jech:
The only hard rule is that I don't use guards (`#ifdef _FOO_H`) around include files, which tend to introduce subtle bugs (it becomes difficult to work out at which point a given include file is being included). The consequence is that I don't `#include""` within `#include""`, since that would require a guard.  I also try to avoid `#include<>` within `#include""`, but I'm not dogmatic about that.

The above implies that header files end up depending on each other.  You may solve the problem either by being careful to include your header files in the right order, or by using forward declarations.  I try to do the former (dependencies between header files), but sometimes it causes either a dependency loop or an explosion in the number of required includes, in which case I use a forward declaration.

In your case, it looks like either approach would do.  The `ubus.h` file is only included in two places, so there's no explosion in the number of includes if you decide on the explicit includes approach, and the `ubus.h` file obviously doesn't need to know about route internals, so a forward declaration is fine.

kerneis:
Thanks a lot for the details, they make sense (and they are indeed a very different approach from the one I'm used to, with header guards). Let's go with forward declarations.

kerneis (approved):

kerneis:
Not merging this right away since there is another pull request, maybe we can review everything and then rebase to make a single package version upgrade.

PolynomialDivision:
Thanks for the explanations. Sounds reasonable.

I never experienced any issues with include-guards, but I think it is very important to follow the same coding style. :)

kerneis:
Can you please change the Makefile to switch to auto-incrementing the package version?

PR #641, opened 2021-02-03
[WIP] babeld: expose add_xroute function via ubus

PolynomialDivision:
If we want to do a Layer3 handover, we need to redirect the traffic to the new AP. However, this does not have to happen for long. A static configuration would probably be too much.

This babel function will allow us to announce routes via ubus.

    ubus call babeld add_xroute '{ "prefix": "2001::1", "plen": 128, "src_prefix": "0::" "src_plen": 0, "metric": 1, "ifindex": 1, "proto": 123 }'

Still thinking about if something like that is really needed since it can be avoided. For now we just add this to show what is possible.

### DIFF
--- a/babeld/src/ubus.c
+++ b/babeld/src/ubus.c
@@ -320,12 +320,78 @@ static int babeld_ubus_get_neighbours(struct ubus_context *ctx_local,
   return ret;
 }
 
+// Enums defining the different ubus-xroute-parameters
+enum {
+  ADD_XROUTE_PREFIX,
+  ADD_XROUTE_PLEN,
+  ADD_XROUTE_SRC_PREFIX,
+  ADD_XROUTE_SRC_PLEN,
+  ADD_XROUTE_METRIC,
+  ADD_XROUTE_IFINDEX,
+  ADD_XROUTE_PROTO,
+  __ADD_XROUTE_MAX
+};
+
+// Json policy describing the ubus-xroute-parameters
+static const struct blobmsg_policy add_xroute_policy[__ADD_XROUTE_MAX] = {
+    [ADD_XROUTE_PREFIX] = {"prefix", BLOBMSG_TYPE_STRING},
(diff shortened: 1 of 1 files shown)


kerneis:
I'm conflicted on this approach, for two reasons:
- How important to you is it to retain json as a configuration language? Babeld already has a language for [configuration changes](https://www.irif.fr/~jch/software/babel/babeld.html#LOCAL_CONFIGURATION_INTERFACE) and it looks like we could be much more generic and avoid duplication by having something like `ubus call babeld configure "..."`?
- it looks like currently babeld does not have an configuration option to manually add an xroute like this. Instead, you specify the kernel routing tables to monitor, and filters to allow or discard some routes. If adding an xroute explicitly is something that makes sense, I think it should be reviewed/added upstream (@jech for input).

jech:
I strongly object to this.  This doesn't guarantee congruence between the data plane and the control plane (babeld is being coerced into announcing routes that don't exist in the data plane), which may cause routing loops or other pathologies.

The proper way is to do redistribution.  Pick an integer, say 13, then add the following to your babeld.conf:

    redistribute proto 13 allow

Then, when a new default route appears, you should do

    ip route add proto 13 ...

and babeld will pick up the route straight away.  When the route disappears, do

    ip route del ...

and babeld will retract the route.

Note that babeld will redistribute unreachable routes, so you don't need to actually install routes that forward traffic, you may install unreachable routes.  The unreachable route will drop traffic if there isn't a higher-priority route, which will prevent babeld from causing a routing loop.

Please see the `babel-pinger` package for an example of this technique.

PolynomialDivision:
Thanks for all the interesting suggestions! You are right. I was to fast with coding.

PolynomialDivision:
> How important to you is it to retain json as a configuration language?

Json is the requirement for ubus communication.

I'm still learning babel. But our network is continiously growing. :)
If u are interested, I hope I can somehow sample statistics with my ubus bindings and provide it via collectd.

PR #642, opened 2021-02-08
babeld: enable ubus by default

PolynomialDivision:
"luci-app-babeld" is currently rewritten in javascript. We use the ubus-api to query all the statistics. That is why we should enable ubus_bindings by default.

https://github.com/openwrt/luci/pull/4779

### DIFF
--- a/babeld/Makefile
+++ b/babeld/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=babeld
 PKG_VERSION:=1.9.2
-PKG_RELEASE:=4
+PKG_RELEASE:=5
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://www.irif.fr/~jch/software/files/
--- a/babeld/files/babeld.config
+++ b/babeld/files/babeld.config
@@ -24,7 +24,7 @@ config general
 	## See comment at the top of this file for more details.
 	# option 'conf_file' '/etc/babeld.conf'
 	# option 'conf_dir' '/tmp/babel.d/'
-	# option 'ubus_bindings' 'false'
+	option 'ubus_bindings' 'true'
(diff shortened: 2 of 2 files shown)


kerneis:
I don't every babeld deployment is using luci for configuration (I'd even go as far as saying I'd expect most people to edit directly babel.conf). Is it possible instead to have a post-install hook to enable ubus when luci-app-babeld is installed?

kerneis:
Another option may be to have a button in LUCI to "enable ubus and reload babeld to see statistics".

PolynomialDivision:
> I don't every babeld deployment is using luci for configuration (I'd even go as far as saying I'd expect most people to edit directly babel.conf). Is it possible instead to have a post-install hook to enable ubus when luci-app-babeld is installed?

Okay, that is also doable. :)

PolynomialDivision:
@kerneis are u more sophisticated with javascript?

I layed here the groundwork how to work with javascript and ubus in babeld:
https://github.com/openwrt/luci/pull/4780

But I think a person with more javascript experience can make a more sophisticated app. ;)

kerneis:
> @kerneis <https://github.com/kerneis> are u more sophisticated with javascript?


Not at all, just ask @jech, he had to maintain my javascript code once and ended up rewriting it from scratch :-D

PolynomialDivision:
> Not at all, just ask @jech, he had to maintain my javascript code once and ended up rewriting it from scratch :-D

Would be awesome! xP

PolynomialDivision:
@JonnyTischbein currently makes a rewrite of my PR. :)
I will write some uci default for the babel app, so I will close.

PolynomialDivision:
Finally. :) https://github.com/openwrt/luci/pull/4791

jech:
> just ask @jech, he had to maintain my javascript code once and ended up rewriting it from scratch :

You're being too modest — as far as I can tell, your code was fine.  But it was server-side node.js code that we wanted to rewrite in Go.

PR #643, opened 2021-02-15
alfred: Fix state handling for startup and restart/reload of alfred

ecsv:
@simonwunderlich 

---

Alfred can be stopped during a (re)start or reload when:

* disabled = 1
* batman-adv interface is missing
* MAC based EUI64 IPv6 link-local address not set

It is assumed that procd detects that the procd instance definition disappeared and the currently running instance must be stopped. But all these checks were triggering an "exit" which then stopped the alfred init script and the wrapper from /lib/functions/procd.sh. And thus procd was never informed about the instances which should be changed.

The correct handling is to return the result of the function instead of directly killing the init scripts.

Fixes: 45db0e60d2ff ("alfred: use procd")

----

The alfred server always needs interfaces to operate on. But these interfaces might not exist at the moment when the daemon process is started. This situation stopped the startup process after the init scripts waited for a longer period of polling the system state.

But alfred is able to deal with interfaces which disappeared at runtime but existed at startup. To force a similar behavior for the alfred startup, the parameter "--force" or "-f" is used. The extra polling code is therefore no longer needed in the init scripts.

### DIFF
--- a/alfred/Makefile
+++ b/alfred/Makefile
@@ -4,7 +4,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=alfred
 PKG_VERSION:=2021.0
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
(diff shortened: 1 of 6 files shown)


adschm:
@simonwunderlich This is now missing in openwrt-21.02 branch, as this has been created slightly before your merge.

PR #684, opened 2021-06-12
oonf: fix compilation now that Ninja is in the tree.

neheb:
ping @mwarning

### DIFF
--- a/oonf-dlep-proxy/Makefile
+++ b/oonf-dlep-proxy/Makefile
@@ -16,18 +16,18 @@ CMAKE_INSTALL:=1
 include $(INCLUDE_DIR)/package.mk
 include $(INCLUDE_DIR)/cmake.mk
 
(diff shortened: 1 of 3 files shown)


PolynomialDivision:
@neheb Could you explain for me why it fixes the compilation?

PolynomialDivision:
```
 Package HASH check failed
 ```

neheb:
note that I haven't touched hashes.

It's needed because cmake.mk sets CMAKE_GENERATOR with a +. See https://github.com/openwrt/openwrt/commit/97258f53634d7237a2962aec3387f011047ce83b

Also see similar fixes: https://github.com/openwrt/openwrt/commit/09de28090cfb3fb605e6f4b452503a4ec2c6c43f

PolynomialDivision:
There is no PKG_HASH at all. xD

PR #690, opened 2021-06-17
issuetemplate: reference foreign repos

SvenRoederer:
As we are the routing-feed, we don't care for packages of the base-, luci- and packages-feed.
So update the imported issue-tempalte accordingly


PolynomialDivision:
I'm unsure about the prefix, but otherwise lgtm.

mwarning:
Maybe use `CI:` as prefix for the commit message?

PolynomialDivision:
> Maybe use `CI:` as prefix for the commit message?

Hmmmm. I do not know. Issuetemplate is also fine for me. Maybe we just merge?

mwarning:
agreed

PR #692, opened 2021-06-18
olsrd: fix olsrd starting before network

### LINKED ISSUE #691: olsrd: not starting with procd - as no interface configured
Environment: OpenWrt-21.02.0-rc3; TPlink WDR3600

Description: 

I recently build a image with all changes from OpenWrt-21.02.0-rc2 to OpenWrt-21.02.0-rc3 (which includes the recent changes of olsrd). With only these changes olsrd fails to start initially, but after manually restarting it later it's working.
In the attached system-log there is a message about `olsrd exit: main: Bad configuration`. 
In the attached logfiles you can see, that during 1st-start (bootup) there is no interface defined. In the config after the 2nd-start (manually triggered) both expected interfaces are defined.
The parallel configuration of the olsrd6 is working fine in the 1st start.

Looks like a race condition with bringing up the interfaces.

[system-log.txt](https://github.com/openwrt/routing/files/6672337/system-log.txt)
[olsr_1ststart.conf.txt](https://github.com/openwrt/routing/files/6672347/olsr_1ststart.conf.txt)
[olsr_2ndstart.conf.txt](https://github.com/openwrt/routing/files/6672348/olsr_2ndstart.conf.txt)

PolynomialDivision:
Should fix #691.

### DIFF
--- a/olsrd/files/olsrd6.init
+++ b/olsrd/files/olsrd6.init
@@ -10,6 +10,17 @@ OLSRD=olsrd6
 CONF=/var/etc/$OLSRD.conf
 PID=/var/run/olsrd6.pid
 
+wait_for_network()
+{
+	ubus -t 15 wait_for network.interface.$1 2>/dev/null
+}
+
+boot()
+{
+	config_load network
+	config_foreach wait_for_network interface
+	rc_procd start_service
+}
 
 start_service() {
 	olsrd_generate_config $OLSRD
--- a/olsrd/files/olsrd4.init
+++ b/olsrd/files/olsrd4.init
@@ -10,6 +10,18 @@ OLSRD=olsrd
 CONF=/var/etc/$OLSRD.conf
 PID=/var/run/olsrd.pid
 
+wait_for_network()
+{
+	ubus -t 15 wait_for network.interface.$1 2>/dev/null
+}
+
+boot()
+{
+	config_load network
+	config_foreach wait_for_network interface
+	rc_procd start_service
+}
+
 start_service() {
 	olsrd_generate_config $OLSRD
 	


olsrd/files/olsrd6.init
```diff
@@ -10,6 +10,17 @@ OLSRD=olsrd6
 CONF=/var/etc/$OLSRD.conf
 PID=/var/run/olsrd6.pid
 
+wait_for_network()
```
SvenRoederer, line 13: The wait_for_network() function is identical to the one in olsrd4.init. Think it should be put into lib/functions/olsrd.sh

same might be usefull for the body of boot() too

SvenRoederer:
* Before merging the message should be changed to be more verbose and explain what's going on than just referencing an issue.
* The solution adds a delay during the service startup, which should work for the #692 case. 
  * But is it delaying the general startup of the system till the interface if up or the timeout is reached? Or will just the olsrd start delayed, but other services continue starting up in parallel
  * What happens to the next type of interface, that just needs 15.3 secs to come up?

PolynomialDivision:
Could you please try disabling ipv6?

Noki:
I tried this fix and it solved my issue, that olsr deamons were started before the interface was ready. Works for olsrd (IPv4) and olsrd6 (IPv6). Please merge this fix!

olsrd/files/olsrd6.init
PolynomialDivision: I'm unsure, I will merge this fix first, since it was approved and then tidy up again.

PolynomialDivision: feel free to open a pr

PolynomialDivision:
> I tried this fix and it solved my issue, that olsr deamons were started before the interface was ready. Works for olsrd (IPv4) and olsrd6 (IPv6). Please merge this fix!

Thakns I will do!

SvenRoederer:
it's solving the issue, but with some things to keep in mind:
* it will likely fail on links taking more than 15 seconds to come up
* it extends the bootup time from 60s to 82s (>33%) in my case

PR #694, opened 2021-06-20
olsrd: add olsrd-neigh.sh

Akira25:
This script originates from Freifunk Berlin. It prints a list of
all meshing neighbors known to olsrd on the command line.

Signed-off-by: Martin Hübner <martin.hubner@web.de>

Description:

### DIFF
--- a/olsrd/Makefile
+++ b/olsrd/Makefile
@@ -157,6 +157,12 @@ define Package/olsrd-mod-watchdog
   TITLE:=Watchdog plugin
 endef
 
+define Package/olsrd-utils
+  $(call Package/olsrd/template)
+  DEPENDS:=olsrd
+  TITLE:=Utils for OLSRD
+endef
+
 define Package/olsrd-mod-pud/conffiles
 /etc/olsrd.d/olsrd.pud.position.conf
 endef
@@ -298,6 +304,11 @@ define Package/olsrd-mod-watchdog/install
 	$(INSTALL_BIN) $(PKG_BUILD_DIR)/lib/watchdog/olsrd_watchdog.so.* $(1)/usr/lib/
 endef
 
+define Package/olsrd-utils/install
+	$(INSTALL_DIR) $(1)/usr/bin
+	$(INSTALL_BIN) ./files/olsrd-neigh.sh $(1)/usr/bin/olsrd-neigh
+endef
+
 $(eval $(call BuildPackage,olsrd))
 $(eval $(call BuildPackage,olsrd-mod-arprefresh))
 $(eval $(call BuildPackage,olsrd-mod-dot-draw))
@@ -318,3 +329,4 @@ $(eval $(call BuildPackage,olsrd-mod-secure))
 $(eval $(call BuildPackage,olsrd-mod-sgwdynspeed))
 $(eval $(call BuildPackage,olsrd-mod-txtinfo))
 $(eval $(call BuildPackage,olsrd-mod-watchdog))
+$(eval $(call BuildPackage,olsrd-utils))
--- a/olsrd/files/olsrd-neigh.sh
+++ b/olsrd/files/olsrd-neigh.sh
@@ -1,0 +1,109 @@
+#!/bin/sh
+
+. /usr/share/libubox/jshn.sh
+
+hostsfile_getname()
+{
+	local config="$1"
+	local i=0
+	local value file
+
+	while value="$( uci -q get $config.@LoadPlugin[$i].library )"; do {
+		case "$value" in
+			'olsrd_nameservice'*)
+				file="$( uci -q get $config.@LoadPlugin[$i].hosts_file )"
+				break
+			;;
+		esac
+
+		i=$(( i + 1 ))
+	} done
+
+	echo "${file:-/var/run/hosts_olsr}"
+}
+
+read_hostnames()
+{
+	local file_list=" $( hostsfile_getname 'olsrd' ) $(hostsfile_getname 'olsrd6' ) "
+	local line ip hostname file file_list_uniq
+
+	for file in $file_list; do {
+		case " $file_list_uniq " in
+			*" $file "*)
+			;;
+			*)
+				file_list_uniq="$file_list_uniq $file"
+			;;
+		esac
+	} done
+
+	for file in $file_list_uniq; do {
+		[ -e "$file" ] || continue
+
+		while read -r line; do {
+			case "$line" in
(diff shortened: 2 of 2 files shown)


olsrd/files/olsrd-neigh.sh
```diff
@@ -0,0 +1,103 @@
...
+	} done
+
+	for file in $file_list_uniq; do {
+		[ -e "$file" ] || continue
+
+		while read -r line; do {
+			case "$line" in
+				[0-9]*)
+					# 2001:bf7:820:901::1 stuttgarter-core.olsr   # myself
+					# 10.63.160.161  AlexLaterne    # 10.63.160.161
```
PolynomialDivision, line 47: Please remove those comments.

olsrd/files/olsrd-neigh.sh
```diff
@@ -0,0 +1,103 @@
...
+					set -f
+					set +f -- $line
+					ip="$1"
+					hostname="$2"
+
+					# global vars, e.g.
+					# IP_1_2_3_4='foo' or IP_2001_bf7_820_901__1='bar'
```
PolynomialDivision, line 54: please remove this line or do not use a public ipv6 as ocmment

olsrd/files/olsrd-neigh.sh
```diff
@@ -0,0 +1,103 @@
...
+}
+
+read_hostnames
+
+VARS='localIP:Local remoteIP:Remote validityTime:vTime linkQuality:LQ'
+VARS="$VARS neighborLinkQuality:NLQ linkCost:Cost remoteHostname:Host"
+
+for HOST in '127.0.0.1' '::1';do
+	json_init
+	json_load "$( echo /links | nc $HOST 9090 | sed -n '/^[}{ ]/p' )"	# remove header/non-json
```
PolynomialDivision, line 69: Please check for erros. A lot of people are using ipv4 olsr only. This will result that ::1:9090 is not reachable and the erorr message looks like this:

```
nc: can't connect to remote host: Connection refused
Failed to parse message data
```

olsrd/Makefile
```diff
@@ -193,6 +193,8 @@ define Package/olsrd/install
 	$(INSTALL_DIR) $(1)/etc/init.d
 	$(INSTALL_BIN) ./files/olsrd4.init $(1)/etc/init.d/olsrd
 	$(INSTALL_BIN) ./files/olsrd6.init $(1)/etc/init.d/olsrd6
+	$(INSTALL_DIR) $(1)/usr/bin
+	$(INSTALL_DATA) ./files/olsrd-neigh.sh $(1)/usr/bin/olsrd-neigh.sh
```
PolynomialDivision, line 197: I would suggest you make a seperate package and call it olsrd-utils-neighbors or something like this.

Akira25: I don't agree. A separate package for just one script seems looks like overkill to me. Maybe we should postpone that new package until there are additional scripts.

olsrd/files/olsrd-neigh.sh
Akira25: Why? They show the structure of the data that is processed. In my understanding, they help to get the gist of that lines.

olsrd/files/olsrd-neigh.sh
Akira25: Resolved in new version.

olsrd/files/olsrd-neigh.sh
Akira25: The comment on the foregoing comment applies here too. :)

olsrd/files/olsrd-neigh.sh
PolynomialDivision: I wouldn't publish personal information in such a script. I would prefer constructing random ip addresses.

olsrd/files/olsrd-neigh.sh
```diff
@@ -0,0 +1,109 @@
...
+}
+
+read_hostnames
+
+VARS='localIP:Local remoteIP:Remote validityTime:vTime linkQuality:LQ'
+VARS="$VARS neighborLinkQuality:NLQ linkCost:Cost remoteHostname:Host"
+
+for HOST in '127.0.0.1' '::1';do
+	# check for availability of HOST
+	ping $HOST -c1 >/dev/null 2>&1
```
PolynomialDivision, line 69: why ping? you could also use the nc command for that?

Akira25: I've tested this code on my router and it's reliably working. If you really need to do it with `nc` free to do it that way instead. Gives pretty much the same result.

olsrd/Makefile
```diff
@@ -298,6 +304,11 @@ define Package/olsrd-mod-watchdog/install
 	$(INSTALL_BIN) $(PKG_BUILD_DIR)/lib/watchdog/olsrd_watchdog.so.* $(1)/usr/lib/
 endef
 
+define Package/olsrd-utils/install
+	$(INSTALL_DIR) $(1)/usr/bin
+	$(INSTALL_DATA) ./files/olsrd-neigh.sh $(1)/usr/bin/olsrd-neigh.sh
```
PolynomialDivision, line 309: You have to istall that as binary. INSTALL_BIN. Maybe also just rename that in /usr/bin/olsrd-neigh

Akira25:
@PolynomialDivision fully working now. ready for merge.

@mwarning

PolynomialDivision:
> @PolynomialDivision fully working now. ready for merge.
> 
> @mwarning

Nice! I will immidiately test and if it is working, also backport to 21.02

olsrd/Makefile
pmelange: I think it is better to use the name we are all familiar with neigh.sh

PolynomialDivision:
Works perfectly! Thanks for your contribution.

SvenRoederer:
> This script originates from Freifunk Berlin. 

To be correct, this code originates from LuCI, based on work of @mmunz, @stargieg, @bittorf and @booo. A reference to the original history might be also nice for people interested in it. (https://github.com/freifunk/openwrt-packages/commits/5f65547dfd59ae3fe9e45f0ac4bc4a06808b4847/contrib/package/freifunk-common/files/usr/bin/neigh.sh). 
Imho it's always a nightmare to see code just falling from the sky, an the main reason why repositories are used.


It would also be great to drop a note to upstream / origin that the code can be removed there and not have to be carried around 2 times.

PR #697, opened 2021-07-02
opennds: Release v9.1.0

bluewavenet:
Maintainer: Rob White <rob@blue-wave.net>
Compile tested: arm_cortex-a7_neon-vfpv4, mipsel_24kc
Run tested: arm_cortex-a7_neon-vfpv4, mipsel_24kc, gl-inet b1300, gl-inet mt300n-v2, Snapshot, 21.02.0-rc3, 19.07.7

Description:
This version introduces new functionality, some changes and fixes
Tested on OpenWrt, OpenSuse and Debian.

From the Changelog:
  * Add - option statuspath to enable alternate status page scripts [bluewavenet]
  * Add - ndsctl lockf() file locking [bluewavenet] [T-X]
  * Add - b64encode to ndsctl [bluewavenet]
  * Add - option max_page_size for MHD [bluewavenet]
  * Add - option remotes_refresh_interval [bluewavenet]
  * Add - Pre-download remote files in background after startup [bluewavenet]
  * Add - client id data files created by openNDS on client connect [bluewavenet]
  * Add - check routing is configured and up [bluewavenet]
  * Add - support for Preemptive Authentication for connected client devices. [bluewavenet]
  * Add - Gateway interface watchdog [bluewavenet]
  * Remove - deprecated IFB config [bluewavenet]
  * Fix - ndsctl, send return codes [bluewavenet]
  * Fix - MHD Watchdog Use uclient-fetch in OpenWrt [bluewavenet]
  * Fix - Improve MHD watchdog [bluewavenet]
  * Fix - update legacy code in ndsctl_thread [bluewavenet]
  * Fix - edge case where MHD returns (null) as host value [bluewavenet]

Signed-off-by: `Rob White <rob@blue-wave.net>`

### DIFF
--- a/opennds/Makefile
+++ b/opennds/Makefile
@@ -7,12 +7,12 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=opennds
 PKG_FIXUP:=autoreconf
-PKG_VERSION:=9.0.0
+PKG_VERSION:=9.1.0
 PKG_RELEASE:=1
 
 PKG_SOURCE_URL:=https://codeload.github.com/opennds/opennds/tar.gz/v$(PKG_VERSION)?
 PKG_SOURCE:=opennds-$(PKG_VERSION).tar.gz
-PKG_HASH:=466dbd78a9464d56f6bf8c8374071e4c21854700a8176a47e72e23930e4271b2
+PKG_HASH:=6064b906b820fb960b15321e83708bd074e1bb797f7741d73740802ad94f7b0c
 PKG_BUILD_DIR:=$(BUILD_DIR)/openNDS-$(PKG_VERSION)
 
 PKG_MAINTAINER:=Rob White <rob@blue-wave.net>


mwarning:
@PolynomialDivision do you why the formalities check fails despite the commit message being fixed?

PolynomialDivision:
I would suggest to use something like `opennds: update to v9.1.0` since `AUTORELEASE` is checking for such stuff.
Your commit fails because the author is `rob <...>`. Please use

    git commit --amend --author="... <your@email.com"
    
and force push again.

Further the glibc build is failing.

mwarning:
The email is actually set to `Rob White <rob@blue-wave.net>`, so that looks fine. Or do I miss somethinng here?

PolynomialDivision:
> The email is actually set to `Rob White <rob@blue-wave.net>`, so that looks fine. Or do I miss somethinng here?

It is not the email, it is the name of the author of the commit message. ;) it is set to rob

mwarning:
foo

I do not see it. :]

Akira25:
@mwarning It's that one that fails:

image

bluewavenet:
@mwarning @PolynomialDivision 
Can "Author:" be set in the commit message in the same way "Maintainer:" and "Signed-off-by:" seem to be set, or do we need to do the:
`git commit --amend --author="... <your@email.com"`

Presumably the current setting for author is coming from my local git repository - a default I guess as I have not deliberately set it.

The commit message method, if it works, would be the most flexible in my opinion as it would override any "random" setting people may have in their git setup.

Re the Test Build / Test arc_archs failure, it is interesting the warnings and failure reported there do not show up when I compile using the SDK. The binaries I created and tested run with no problems so the error found is not critical in any way **_in this case_**.

In my opinion these "new" tests are very worthwhile not least from a quality control point of view.

If we can resolve the Formalities issue here I will probably close this PR, fix the Test Build / Test arc_archs problems at source then do a new release before coming back here with a new PR.

PolynomialDivision:
> Presumably the current setting for author is coming from my local git repository - a default I guess as I have not deliberately set it.

Yep. ;)



> Re the Test Build / Test arc_archs failure, it is interesting the warnings and failure reported there do not show up when I compile using the SDK. The binaries I created and tested run with no problems so the error found is not critical in any way **_in this case_**.

Did you also use glibc? This target is the only one using glibc. However, I think it is also fine to merge and fix glibc later, since I assume opendns is not building for this target currently.

bluewavenet:
@PolynomialDivision 

Yes, but:

> Can "Author:" be set in the commit message in the same way "Maintainer:" and "Signed-off-by:" seem to be set, or do we need to do the:
> `git commit --amend --author="... <your@email.com"`

?

bluewavenet:
@PolynomialDivision 

> Did you also use glibc?

Yes, version 2.26-lp152.26.3.1

bluewavenet:
@PolynomialDivision @mwarning 
In answer to my own question, the script embedded in formal.yml does not allow override of "Author:" in the commit message, instead it takes the first occurrence of "Author:" it finds.

In formal.yml, it currently has:
```
            author="$(git show -s --format=%aN $commit)"
            if echo $author | grep -q '\S\+\s\+\S\+'; then
              success "Author name ($author) seems ok"
            else
              err "Author name ($author) need to be your real name 'firstname lastname'"
              RET=1
            fi

```
If instead it had something like:

```
            author=$(git show -s $commit)
            author_parsed=$(echo $author | awk -F"Author: " '{print $2"\n"$3"\n"}' | awk -F"> " 'NF>0 {print $1">"}' | grep "\S\+\s\+\S\+ <")

            if [ ! -z "$author_parsed" ]; then
              success "Author name ($author) seems ok"
            else
              err "Author name ($author) need to be your real name 'firstname lastname'"
              RET=1
            fi
```
it would allow override.

Of course it is no big deal for me to set `git commit --amend --author="...` but an override would give flexibility and save time, particularly where third party (corporate, company etc) standards clash with the formalities check...

Just thinking out loud...

bluewavenet:
Closing this PR.
Will release openNDS 9.1.1 with fixups for the errors warnings identified here then open a new PR.

PR #702, opened 2021-07-13
naywatch: add naywatch

PolynomialDivision:
The network may crash on the SoCs, but the SoC itself does not. This leads to a node no longer being accessible in a mesh network.  If the node is placed in a location that is not easily accessible, e.g. in a high tower, it can cause a lot of problems. Therefore we check the link-local connectivity on the configured interfaces.

### DIFF
--- a/naywatch/files/naywatch.init
+++ b/naywatch/files/naywatch.init
@@ -1,0 +1,65 @@
+#!/bin/sh /etc/rc.common
+
+USE_PROCD=1
+START=95
+STOP=01
+
+log() {
+    local msg="$1"
+    logger -t naywatch "$msg"
+}
+
+wait_for_network()
+{
+    ubus -t 15 wait_for network.interface.$1 2>/dev/null
+}
+
+boot()
+{
+    local _interfaces
+    config_load naywatch
+    config_get _interfaces general interface
+
+    for interface in $_interfaces; do
+        wait_for_network interface
+    done
+
+    rc_procd start_service
+}
+
+start_service() {
+    procd_open_instance
+
+    config_load naywatch
+    local _check_interval
+    local _watchdog_timeout
+    local _use_watchdog
+    local _save_logs
+    local _interfaces
+
+    config_get _check_interval general "check_interval"
+    config_get _watchdog_timeout general "watchdog_timeout"
+    config_get _use_watchdog general "use_watchdog"
+    config_get _save_logs general "save_logs"
+    config_get _interfaces general "interface"
+
+    procd_set_param command /usr/bin/naywatch "$_check_interval" "$_watchdog_timeout" "$_use_watchdog" "$_save_logs" "$_interfaces"
+
+    procd_set_param respawn 3600 15 0
+
+    procd_set_param stdout 1
+    procd_set_param stderr 1
+
+    procd_close_instance
+}
+
+stop_service() {
+    exec 3>&- # close file again
+    sync && wait
+}
+
+service_stopped() {
+    log "Naywatch Stopped!"
+    log "Handover Watchdog to procd again:"
+    ubus call system watchdog '{"magicclose":true,"stop":false}' > /dev/null
+}
--- a/naywatch/files/naywatch.sh
+++ b/naywatch/files/naywatch.sh
@@ -1,0 +1,117 @@
+#!/bin/sh
+
+. /lib/functions.sh
+. /lib/functions/network.sh
+
+CHECK_INTERVAL=$1
+shift
+WATCHDOG_TIMEOUT=$1
+shift
+USE_WATCHDOG=$1
+shift
+SAVE_LOGS=$1
+shift
+INTERFACES="$*"
+
+ACTIVE=0
+
+log() {
+    local msg="$1"
+    logger -t naywatch "$msg"
+}
+
+write_logs() {
+    save_log() {
+        eval $1 > /root/$(date +%s)-"$1".log
+    }
+    config_load naywatch
+    config_list_foreach general save_cmd save_log
+    sync
+}
+
+neighbors_available() {
+    local phy
+
+    for interface in $INTERFACES; do
+        network_get_physdev phy $interface > /dev/null 2>&1
+        linklocal=$(ip -6 a list dev $phy | grep "scope link" | awk '{print $2}' | sed 's/\/64//') 2> /dev/null
+        ips=$(ping ff02::1%$phy -w5 -W5 -c2 | awk '/from/{print($4)}' | sed 's/.$//') 2> /dev/null
+        for ip in $ips; do
+            if [ $ip != $linklocal ] && [ $(owipcalc $ip linklocal) -eq 1 ]; then
+                echo 1
+                return 0
+            fi
+        done
+    done
+
+    echo 0
+}
+
+activate_watchdog() {
+    # disable openwrt instrumentation:
+    ubus call system watchdog '{"magicclose":true,"stop":true,"timeout":'${WATCHDOG_TIMEOUT}'}' > /dev/null
+    exec 3>/dev/watchdog
+}
+
+reboot_now() {
+    # copied from watch-cat
+    reboot &
+
+    [ "$1" -ge 1 ] && {
+        sleep "$1"
+        echo 1 >/proc/sys/kernel/sysrq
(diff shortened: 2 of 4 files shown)


mesh-watcher/files/mesh-watcher.sh
```diff
@@ -0,0 +1,50 @@
...
+    config_load babeld
+    mesh_interfaces=$(config_foreach babeld_get_mesh_interface interface ifname)
+
+    for mesh in $mesh_interfaces; do
+        network_get_physdev phy $mesh
+        linklocal=$(ip -6 a list dev $phy | grep "scope link" | awk '{print $2}' | sed 's/\/64//')
+        ips=$(ping ff02::1%$phy -c2 | grep from | awk '{print $4}' | sed 's/.$//')
+        for ip in $ips; do
+            if [ $ip != $linklocal ] && [ $(owipcalc $ip linklocal) -eq 1 ]; then
+                echo $ip
```
Noki, line 23: You can break the loops once you found a single one and save some CPU time by ending early. Maybe also rename the function to "has_neighbors".

mesh-watcher/Makefile
```diff
@@ -0,0 +1,45 @@
...
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=mesh-watcher
+PKG_VERSION:=1
+PKG_RELEASE:=$(AUTORELEASE)
+
+PKG_MAINTAINER:=Nick Hainke <vincent@systemli.org>
+PKG_LICENSE:=GPL-2.0
```
adschm, line 13: Should probably be GPL-2.0-only like on the top.

SvenRoederer:
I read this script like: "my node directly reboots, when my only neighbour turns of his node". Is this correct?

PolynomialDivision:
> I read this script like: "my node directly reboots, when my only neighbour turns of his node". Is this correct?

@SvenRoederer I changed that now to whatever interface you like. ;) For example you can use mgmt.

spolack:
Instead of waiting until network interfaces are up, i propose u to remain in dry-run mode, until the ping succeeded for the first time.

spolack:
Furthermore a list of commands which output will be collected before reboot should be configurable.

PolynomialDivision:
> Instead of waiting until network interfaces are up, i propose u to remain in dry-run mode, until the ping succeeded for the first time.

maybe both?

spolack:
> > Instead of waiting until network interfaces are up, i propose u to remain in dry-run mode, until the ping succeeded for the first time.
> 
> maybe both?

whats the advantage?

PolynomialDivision:
> Furthermore a list of commands which output will be collected before reboot should be configurable.

probably we could give some path to a shell script which should be called?

PolynomialDivision:
> > > Instead of waiting until network interfaces are up, i propose u to remain in dry-run mode, until the ping succeeded for the first time.
> > 
> > 
> > maybe both?
> 
> whats the advantage?

That the network is correctly setup before starting the daemon. I don't know. Just feels saver to me.

spolack:
> > Furthermore a list of commands which output will be collected before reboot should be configurable.
> 
> probably we could give some path to a shell script which should be called?

whats wrong with a list in uci?

PolynomialDivision:
> > > Furthermore a list of commands which output will be collected before reboot should be configurable.
> > 
> > 
> > probably we could give some path to a shell script which should be called?
> 
> whats wrong with a list in uci?

Added.

PolynomialDivision:
I renamed package to naywatch, since it is a cool name. ;)

Noki (approved):
Thanks for doing the changes. Looks good now!

PolynomialDivision:
@mwarning @adschm Anything left? Otherwise, I would like to merge and add this tool also to OpenWrt 21.02.

naywatch/files/naywatch.sh
```diff
@@ -0,0 +1,115 @@
...
+    config_load naywatch
+    config_list_foreach general save_cmd save_log
+    sync
+}
+
+neighbors_available() {
+    for interface in $INTERFACES; do
+        network_get_physdev phy $interface > /dev/null 2>&1
+        linklocal=$(ip -6 a list dev $phy | grep "scope link" | awk '{print $2}' | sed 's/\/64//') 2> /dev/null
+        ips=$(ping ff02::1%$phy -c2 | grep from | awk '{print $4}' | sed 's/.$//') 2> /dev/null
```
mwarning, line 36: This could be done shorter.
`grep from | awk '{print $4}'`
=>
`awk '/from/{print($4)}'`
(same for the other calls)
The sed call can be included using `awk -F '[. ]+' ...`).

The ping calls might benefit from some timeout/deadline (-W, -w).

naywatch/files/naywatch.init
```diff
@@ -0,0 +1,65 @@
...
+    local _save_logs
+    local _interfaces
+
+    config_get _check_interval general check_interval
+    config_get _watchdog_timeout general watchdog_timeout
+    config_get _use_watchdog general use_watchdog
+    config_get _save_logs general save_logs
+    config_get _interfaces general interface
+
+    procd_set_param command /usr/bin/naywatch $_check_interval $_watchdog_timeout $_use_watchdog $_save_logs $_interfaces
```
mwarning, line 46: `procd_set_param command /usr/bin/naywatch $_check_interval $_watchdog_timeout $_use_watchdog $_save_logs $_interfaces`
=>
`procd_set_param command /usr/bin/naywatch "$_check_interval" "$_watchdog_timeout" "$_use_watchdog" "$_save_logs" "$_interfaces"`
.. might be better, in case one of the arguments is not set.

naywatch/files/naywatch.init
```diff
@@ -0,0 +1,65 @@
...
+    procd_open_instance
+
+    config_load naywatch
+    local _check_interval
+    local _watchdog_timeout
+    local _use_watchdog
+    local _save_logs
+    local _interfaces
+
+    config_get _check_interval general check_interval
```
mwarning, line 40: I usually do `config_get check_interval general "check_interval"`, but that is probably just my personal style. Feel free to ignore. :-)

naywatch/files/naywatch.init
```diff
@@ -0,0 +1,65 @@
...
+
+    procd_set_param stdout 1
+    procd_set_param stderr 1
+
+    procd_close_instance
+}
+
+stop_service() {
+    exec 3>&- # close file again
+    sync & wait
```
mwarning, line 58: What is the purpose if the `&` here?

naywatch/files/naywatch.sh
```diff
@@ -0,0 +1,115 @@
...
+    save_log() {
+        eval $1 > /root/$(date +%s)-"$1".log
+    }
+    config_load naywatch
+    config_list_foreach general save_cmd save_log
+    sync
+}
+
+neighbors_available() {
+    for interface in $INTERFACES; do
```
mwarning, line 33: Add `local phy`?

naywatch/files/naywatch.sh
PolynomialDivision: I tried
```
 ping ff02::1%eth0.42 -w5 -W5 -c2 | awk -F '[. ]+' '/from/{p
rint($4)}'
```
But this awk -F does not work? Just searching for some documentation

naywatch/files/naywatch.init
PolynomialDivision: yeah you are right

naywatch/files/naywatch.init
PolynomialDivision: Changed that. ;)

naywatch/files/naywatch.sh
PolynomialDivision: added

naywatch/files/naywatch.init
PolynomialDivision: You are totally right :)

naywatch/files/naywatch.sh
mwarning: -F sets the regex that is used to separate the field (usually whitespace).
I cannot test right now. Feel free to use or not. :)

PolynomialDivision:
@mwarning Thanks! :)

PR #703, opened 2021-07-13
olsrd: improve respawn settings

Noki:
This increases the amount of seconds to wait before a service restart attempt from 5 to 15 seconds and allows unlimited retries. Olsrd sometimes crashes together with a network interface and 5 seconds can be too short to bring back up a crashed network interface, which is required for olsrd to restart.

Signed-off-by: Tobias Schwarz <info@tobias-schwarz.com>

### DIFF
--- a/olsrd/files/olsrd.config
+++ b/olsrd/files/olsrd.config
@@ -22,3 +22,7 @@ config LoadPlugin
 
 config Interface
 	list interface 'wlan'
+
+config procd general
+	option respawn_timeout '15'
+	option respawn_retry '0'
--- a/olsrd/files/olsrd4.init
+++ b/olsrd/files/olsrd4.init
@@ -24,15 +24,23 @@ boot()
 
 start_service() {
 	olsrd_generate_config $OLSRD
-	
+
 	procd_open_instance
+
+	config_load olsrd
+	local _respawn_timeout
+	local _respawn_retry
+	
+	config_get _respawn_timeout general respawn_timeout	
+	config_get _respawn_retry general respawn_retry		
+	
 	procd_set_param command "$BIN"
 	procd_append_param command -f ${CONF}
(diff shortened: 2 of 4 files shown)


pmelange:
I like the idea of not giving up after just 5 tries.  It would be nice if it were configurable in uci or olsrd.conf

Akira25:
@PolynomialDivision  This looks like a really good feature to me. Could you please review and merge that?

PolynomialDivision:
lgtm

pmelange:
NACK!

https://openwrt.org/docs/guide-developer/config-scripting#reading_options_config_get

What happens if there is no value set for olsrd[6].general.respawn_timeout or respawn_retry in the config files?  I would prefer to have default values in the init scripts in case the options are not set (the optional 4th option to config_get).

```
config_load olsrd

config_get _respawn_timeout general respawn_timeout 15
config_get _respawn_retry general respawn_retry 0
```

pmelange:
Also, I think naming the procd section "general" is unclear.  Better would be 

```
config procd 'procd'
```

PolynomialDivision:
> NACK!
> 
> https://openwrt.org/docs/guide-developer/config-scripting#reading_options_config_get
> 
> What happens if there is no value set for olsrd[6].general.respawn_timeout or respawn_retry in the config files? I would prefer to have default values in the init scripts in case the options are not set (the optional 4th option to config_get).
> 
> ```
> config_load olsrd
> 
> config_get _respawn_timeout general respawn_timeout 15
> config_get _respawn_retry general respawn_retry 0
> ```

Feel free to open PR.

PolynomialDivision:
@pmelange I added your wished changes in latest commit sets.

pmelange:
You were a lot faster that me.  Thanks.  Looks good.

PR #712, opened 2021-08-11
bird2: Fix bus error on OSPF on IPQ806X

mattreeve:
Maintainer: @tohojo 
Compile tested: IPQ806X, Netgear Nighthawk X4S R7800, 21.02-rc4
Run tested: IPQ806X, Netgear Nighthawk X4S R7800, 21.02-rc4. Bird2 now starts when OSPF v2 or v3 is configured.

Description:

On a Netgear R7800, if ospf v2 or v3 is configured in bird.conf, it fails to start with this error:

Fri Jun 11 14:41:11 2021 daemon.info bird: Started
Fri Jun 11 14:41:11 2021 kern.err kernel: [ 3500.853248] Alignment trap: not handling instruction f44c0a1f at [<00035848>] Fri Jun 11 14:41:11 2021 kern.alert kernel: [ 3500.853283] 8<--- cut here ---
Fri Jun 11 14:41:11 2021 kern.alert kernel: [ 3500.859363] Unhandled fault: alignment exception (0x801) at 0x007e0624
Fri Jun 11 14:41:11 2021 kern.alert kernel: [ 3500.862443] pgd = 0bbef4fd
Fri Jun 11 14:41:11 2021 kern.alert kernel: [ 3500.868821] [007e0624] *pgd=5d6ca835, *pte=5c40b75f, *ppte=5c40bc7f

The problem is due to a struct not being properly aligned on the ARMv7 architecture.

This patch fixes the problem by adding the "PACKED" macro to the affected struct. Note
that upstream may later fix this in another way, in which case this patch will not be required.

Also compiled and run tested on ATH79, Netgear WNDR3700 v4, 21.02-rc4 for regression purposes.

### DIFF
--- a/bird2/Makefile
+++ b/bird2/Makefile
@@ -8,7 +8,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=bird2
 PKG_VERSION:=2.0.8
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=bird-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=ftp://bird.network.cz/pub/bird
--- a/bird2/patches/000-ospf-bus-error.patch
+++ b/bird2/patches/000-ospf-bus-error.patch
@@ -1,0 +1,11 @@
+--- a/proto/ospf/topology.h
++++ b/proto/ospf/topology.h
+@@ -41,7 +41,7 @@ struct top_hash_entry
+   u8 mode;			/* LSA generated during RT calculation (LSA_RTCALC or LSA_STALE)*/
+   u8 nhs_reuse;			/* Whether nhs nodes can be reused during merging.
+ 				   See a note in rt.c:add_cand() */
+-};
++} PACKED;
+ 
+ 
+ /* Prevents ospf_hash_find() to ignore the entry, for p->lsrqh and p->lsrth */


tohojo (approved):
LGTM, but I'm afraid I don't have permissions to merge in this repository :(

PolynomialDivision:
> LGTM, but I'm afraid I don't have permissions to merge in this repository :(

Should I give you write privilege in this routing feed?

@mattreeve Please fix your signed off email and further refresh the patch. :)
This is no a upstream backport? Otherwise, I like it if the commit message is also included in the patch, explaining what it does. However, this is no requirement. Thanks for fixing! :)

PolynomialDivision:
Any reason only pushing the fix 21.02? What is with snapshot openwrt builds?

tohojo:
Yes, please, commit access would be wonderful!

And no, not a backport, upstream is still mulling over what they want to do about it...

PolynomialDivision:
@tohojo I can not put you into the routing-admin team. I think @aparcar has to do it. However, I have no idea what our rules for applying commit access are. But as active maintainer of bird I think you definitely have to right to get commit access.

Any nack from a member of @openwrt/routing-admin ?

PolynomialDivision:
@mattreeve You still have to refresh your patch. I think it is possible via `make package/bird2/refresh` (without any guarantees).

mattreeve:
@PolynomialDivision Thanks - I hadn't used quilt which I think was the problem. I've had another go, please bear with me until I get this right! :)

PolynomialDivision:
You're welcome. Ask as many questions as you like. :)

I will merge since @tohojo already approved. However, I would have put the fix into master branch and backport the relevant change into 21.02. And later also apply the upstream fix on both branches. ;)

PolynomialDivision:
> @PolynomialDivision @tohojo I'm happy to add new commiters, ideally 1-2 additional committers from the routing repository _approve_ @tohojo work here since I'm not actively monitoring the commits over here.

@mwarning also approved in private chat. ;)

aparcar:
I added @tohojo to the new @openwrt/routing-write group. All current members of @openwrt/routing-admin should consider if they don't feel more comfortable over there to avoid spreading unnecessary admin access within this repository.

New maintainers of routing.git should be added to @openwrt/routing-write

tohojo:
Awesome, thanks! :D

PolynomialDivision:
> However I've no strong preference, we could also forward-port it to master now if that's deemed to be more appropriate.

That is now up to @tohojo ;) However, I would prefer to have a working bird on master! :D

tohojo:
Yeah, I think we might as well apply this patch to master as well. @mattreeve care to open another PR to do that? :)

zorun:
The standard way of doing things in openwrt is to commit to master and then backport to release branches. So yes, please add the patch to master, and next time do it in the proper order ;)

PR #714, opened 2021-08-12
cjdns: fix uci-defaults

champtar:
Maintainer: @wfleurant 
Environment: master / 21.02
Compile tested: NO
Run tested: NO
Description:

Since https://git.openwrt.org/?p=openwrt/openwrt.git;a=commitdiff;h=4b9a67362d70c544b85078b8d5c661f43f7472d9,
uci network config interface sections use 'device' instead of 'ifname'

Even if 'device' seems to always be set by
package/base-files/files/bin/config_generate, we keep the fallbacks
in case someone install the package via opkg and has a non migrated
network config

### DIFF
--- a/cjdns/Makefile
+++ b/cjdns/Makefile
@@ -18,7 +18,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=cjdns
 PKG_VERSION:=v21
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/cjdelisle/cjdns/tar.gz/$(PKG_NAME)-$(PKG_VERSION)?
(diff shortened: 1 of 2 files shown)


wfleurant:
Let me catch up to just test each case... thanks in advance

wfleurant:
OK.. tested the logic separately. I see what you mean!
image

This makes sense. I first tested the logic on RPi4 without bridge.  
I think on line +42 we should also make these s/ifname/device changes
```
  # create the network interface
  uci -q batch <<-EOF >/dev/null
    set network.cjdns=interface
    set network.cjdns.ifname=tuncjdns <---- Can you patch ifname here too?
    set network.cjdns.proto=none
EOF
```

champtar:
do we also need to add a 'device' uci section ? or just s/ifname/device is enough from what you see in /etc/config/network ?

champtar:
I did the change anyway, I can add `Tested-by: William Fleurant <meshnet@protonmail.com>` when you give your blessing

wfleurant (approved):

wfleurant:
hehe.. thanks a bunch.
> do we also need to add a 'device' uci section

i tested those changes too, all set. it's safe to just squash && merge now.

BKPepe:
This should be backported to OpenWrt 21.02.

PR #729, opened 2021-10-01
naywatch: fix procd handling

PolynomialDivision:
Sometimes, naywatch can not handover the watchdog to procd again using ubus. We need to call the same ubus command multiple times until procd takes over control again.


spolack (commented):
I'd propose to fix that in procd?

PolynomialDivision:
Maybe the file handle is not closed correctly, so it is my bad. However, for me this workaround fixes the issue.

PolynomialDivision:
Maybe @dangowrt can say something to it. However, I will elaborate this further.

PR #734, opened 2021-10-12
olsrd: wait for wireless to initialize before starting

pmelange:
Wait for wireless to initialize (pending="false") to ensure
that any olsrd(6) configured interfaces and their associated
devices are up.  This provides the init script the ability
to properly add all interface sections to the generated
config file /tmp/etc/olsrd(6).conf

Signed-off-by: Perry Melange <isprotejesvalkata@gmail.com>

### DIFF
--- a/olsrd/files/olsrd4.init
+++ b/olsrd/files/olsrd4.init
@@ -15,10 +15,24 @@ wait_for_network()
 	ubus -t 15 wait_for network.interface.$1 2>/dev/null
 }
 
+wait_for_wireless()
+{
+	local count=0
+	ubus -t 15 wait_for network.wireless 2>/dev/null
+	while [ $count -lt 30 ]; do
+		if  [ "$(ubus call network.wireless status | jsonfilter -e '@[*]["pending"]' | grep true | wc -l)" == "0" ]; then
+			break
+		fi
+		count=$((count+1))
+		sleep 1
+	done
+}
+
 boot()
 {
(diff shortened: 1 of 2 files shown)


pmelange:
Upon boot, on a slow router, the init script's first run happens before the wireless initialization process is finished  The result is that wireless mesh interfaces are not being included into /tmp/etc/olsrd.conf.

the problem was reported in the freifunk-berlin chat and I was able to reproduce it on a CPE210.

zioproto:
I understand the problem. The olsrd config in uci does not use real interface names. Example:

`uci set olsrd4.@Interface[0].interface=lan`

The actual hardware interface names are not ready when olsrd script starts on boot.

Do I understand the problem correctly ?

I see the `ubus` command is called already:
https://github.com/openwrt/routing/blob/master/olsrd/files/olsrd4.init#L15

Why it does not work when it is called here ? https://github.com/openwrt/routing/blob/master/olsrd/files/olsrd4.init#L21

zioproto:
I agree we need your patch, but can you improve the existing `wait_for_network` function ?

zioproto:
For readability  would put the new code into a`wait_for_wireless` function ?

Last question: In your while loop there is no timeout ... you could loop forever. In there a better solution ?

thanks

PolynomialDivision:
LGTM for me. Lets wait for @zioproto

Issue #727, opened 2021-09-25
cjdns: cjdroute Segmentation fault


gimzmoe:
Environment: (put here arch, model, OpenWrt version)
Model: ZyXEL NBG6817
Architecture: ARMv7 Processor rev 0 (v7l)
Firmware Version: OpenWrt 19.07.7 r11306-c4a6851c72 / LuCI openwrt-19.07 branch git-21.044.30835-34e0d65
Kernel Version: 4.14.221
Description:
installed cjdns, cjdroute causes seg fault without any other explaination

```
root@OpenWrt:~# opkg list | grep cjdns
cjdns - 20.2-1

root@OpenWrt:~# cjdroute
Segmentation fault

root@OpenWrt:~# cjdrouteconf get | cjdroute
Segmentation fault
```
Occurs on other tp-link and gl-inet devices too

gimzmoe:
trying 21.02

mwarning:
@wfleurant

gimzmoe:
while 21.02 is quite nice and I'm glad I upgraded, the cjdroute seg fault persists.  No tun0 device

wfleurant:
crazy...no? I'll pull and build keeping symbols and see if gdb sheds some light.

wfleurant:
can you try 
`gdb ./cjdroute -ex 'set follow-fork-mode child' -ex 'run' -ex 'thread apply all bt' -ex 'quit'`
or use strace? dmesg?

jine:
I can confirm we're seeing this issue on GL.iNet B1300, with custom firmware based upon 19.07 (but the cjdns-package is from openwrt master repo, so no modifications there) - CPU: Qualcomm IPQ4028

I can check gdb with those params later this week, as i have another task with one of those units. I remember i also tried 21.02 with the exact same segfault - i found some notes in our internal systems about it i saved:


```
root@xxxxxx-xxxxxx:~# cat /etc/os-release
NAME="xxxxxx"
VERSION="21.02-SNAPSHOT"
ID="openwrt"
ID_LIKE="lede openwrt"
PRETTY_NAME="OpenWrt 21.02-SNAPSHOT"
VERSION_ID="21.02-snapshot"
HOME_URL="https://openwrt.org/"
BUG_URL="https://bugs.openwrt.org/"
SUPPORT_URL="https://forum.openwrt.org/"
BUILD_ID="r0+16120-6bb6901de9"
OPENWRT_BOARD="ipq40xx/generic"
[...]

PR #738, opened 2021-10-13
cjdns: pass for some build warnings

wfleurant:
Maintainer: me
Compile tested: builds for x86_64
Run tested: build only testing

Description: these are cjdns author reviewed changes. may (lead to) fix segfault issue #727. Due to stricter compiler(s) we do not want to error for these warnings. Caused by: ...reading "past the end" of a struct, because the struct is actually a header and the body is of unknown size. Compilers got stricter and this became a warning.

### DIFF
--- a/cjdns/Makefile
+++ b/cjdns/Makefile
@@ -18,7 +18,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=cjdns
 PKG_VERSION:=v21
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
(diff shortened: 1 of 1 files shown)


mwarning:
This looks like it is trying to ignore errors. Thats's a band-aid made out of chillies. :>

cjdelisle:
It will eventually, the cause of the warnings is reading "past the end" of a struct, because the struct is actually a header and the body is of unknown size. Compilers got stricter and this became a warning.

BKPepe:
@PolynomialDivision I think this is required to be backported to OpenWrt 21.02. It will fix https://github.com/openwrt/routing/issues/813

PR #742, opened 2021-10-19
[21.02] babeld: remove unnecessray blob_buf_init calls

PolynomialDivision:
For some reason the buffer is initialized twice before filling it.
This does not break anything but is useless. Reduce the initialization
to only one.

(cherry picked from commit 2cf253b3df2db7830692de645504db294151e5ab)


PolynomialDivision:
Yes. I tested it massively on trunk. Furthermore, @mwarning also clicked on "resolve" for the conversation. It was just a mistake from my side that I initialized the buffer twice. It is just useless and is causing unneeded cpu cycles and memory access. However, maybe the compiler also optimizes it and just remove the call.

PR #743, opened 2021-10-19
minimalist-pcproxy,pimgc,mcproxy,mrd6: include PKG_MIRROR_HASH

cotequeiroz:
Maintainers: @sbyx @fingon @Oryon 
Compile tested: mediatek, openwrt master
Run tested: N/A

Description:
I've ran `make package/{minimalist-pcproxy,pimgc,mcprox,mrd6}/{download,check} FIXUP=1` to add the `PKG_MIRROR_HASH` variable to these packages that were missing it.

Setting the variable avoids recreating the tarball every time the package is rebuilt.  It also avoids rebuilds if you run `make` back-to-back without changing anything.


cotequeiroz:
It looks like the Ci did not agree with my mcproxy hash.  I will investigate it.  Meanwhile, let's see if the rest of the pack works.

PolynomialDivision:
Hey @sbyx @fingon @Oryon
Can you ack?

PolynomialDivision:
I looked at the history of the Makefiles and I think they are now maintained by the "whole routing feed". I will merge.

BKPepe:
Can this be backported to OpenWrt 21.02, please? Should I create a pull request?

PR #753, opened 2021-11-19
Revert "batman-adv: update packages to version 2021.4"

simonwunderlich:
Reverts openwrt/routing#750


BKPepe:
Please always provide the reason, why the commit is reverted.

PR #756, opened 2021-11-24
Enable vtysh for all daemons.

GitArUs:
Running watchquagga does not make any sense without vtysh interface accessible via unix sockets created in /var/run/quagga. Without these sockets watchquagga is unable to check status of daemons, so it assumes they are dead and restarts them periodically, even if they are working fine. Only one configure option is needed to make it work properly.


GitArUs:
Sorry, I was wrong. Forgot to reinstall all quagga packages after enabling vtysh in configuration.

PR #757, opened 2021-11-27
batman-adv: Prevent use from libc headers to not build with BIG_ENDIAN

djselbeck:
Commit 97d35a552ec5b6ddf7923dd2f9a8eb973526acea of musl introduced the
macros __LITTLE_ENDIAN and __BIG_ENDIAN in alltypes.h. These are pulled
into the compilation of batman-adv. This has the side effect that the
function is_multicast_ether_addr of etherdevice.h in Linux kernel is
compiled as the big endian version and so fails to work properly on
little endian devices.

This commit fixes that by preventing to pull in std includes and by
providing the minimal stdarg.h that Linux now provides with recent
versions. The included header file comes from Linux upstream commit
c0891ac15f0428ffa81b2e818d416bdf3cb74ab6.

Signed-off-by: Hendrik Borghorst <hendrikborghorst@gmail.com>

Maintainer: @ecsv , @simonwunderlich 
Compile tested: ramips/mt7621/wac124
Run tested: OpenWRT master @ fb6c34d355e19152550a8347cde13e1c359f1aca 

Batman finds neighbor nodes properly again.

Description:

### DIFF
--- a/batman-adv/Makefile
+++ b/batman-adv/Makefile
@@ -61,6 +61,7 @@ PKG_EXTRA_CFLAGS:= \
 	$(patsubst CONFIG_%, -DCONFIG_%=1, $(patsubst %=y,%,$(filter %=y,$(PKG_EXTRA_KCONFIG)))) \
 
 NOSTDINC_FLAGS = \
+	$(KERNEL_NOSTDINC_FLAGS) \
 	-I$(PKG_BUILD_DIR)/net/batman-adv \
 	-I$(STAGING_DIR)/usr/include/mac80211-backport \
 	-I$(STAGING_DIR)/usr/include/mac80211-backport/uapi \


djselbeck:
This PR fixes recent changes on OpenWRT master that broke batman completely for my ramips (mipsel) device because all mac addresses are considered multicast and all ELP packages are discarded.

https://github.com/openwrt/openwrt/commit/aaec2ad13b18086e2aaf9ca9152827c50e578bc0 that commit updated musl

ecsv:
Shouldn't this be part of the mac80211 package? The batman-adv package is using its compat/backport headers.

djselbeck:
Good question, I can not tell. But batman-adv is also pulling in `stdarg.h` directly in `log.c`

batman-adv/Makefile
```diff
@@ -61,6 +61,7 @@ PKG_EXTRA_CFLAGS:= \
 	$(patsubst CONFIG_%, -DCONFIG_%=1, $(patsubst %=y,%,$(filter %=y,$(PKG_EXTRA_KCONFIG)))) \
 
 NOSTDINC_FLAGS = \
+	-nostdinc \
```
ecsv, line 64: This is definitely something which is missing. But it should actually be:

```
$(KERNEL_NOSTDINC_FLAGS) \
```

ecsv:
> Good question, I can not tell. But batman-adv is also pulling in `stdarg.h` directly in `log.c`

This is not actually the problem. Btw. your change is currently creating more problem than it should because your are using `-nostdinc` directly. This is also the reason why you suddenly need this additional stdarg.h (which should come from a different source)

So without ```$(KERNEL_NOSTDINC_FLAGS) \```, you would get following include path to alltypes.h (which causes the reported problem):

```
/openwrt/staging_dir/toolchain-x86_64_gcc-11.2.0_musl/include/stdarg.h:10,
./include/linux/kernel.h:6,
/openwrt/staging_dir/target-x86_64_musl/usr/include/mac80211-backport/linux/kernel.h:3,
./include/asm-generic/bug.h:20,
/openwrt/staging_dir/target-x86_64_musl/usr/include/mac80211-backport/asm-generic/bug.h:3,
./arch/x86/include/asm/bug.h:93,
./include/linux/bug.h:5,
/openwrt/staging_dir/target-x86_64_musl/usr/include/mac80211-backport/linux/bug.h:3,
/openwrt/staging_dir/target-x86_64_musl/usr/include/mac80211/linux/average.h:5,
/openwrt/build_dir/target-x86_64_musl/linux-x86_64/batman-adv-2021.1/compat-hacks.h:55,
<command-line>:
```

With ```$(KERNEL_NOSTDINC_FLAGS) \```, you would get following include path to stdarg.h (which doesn't include alltypes.h):

```
/staging_dir/toolchain-x86_64_gcc-11.2.0_musl/lib/gcc/x86_64-openwrt-linux-musl/11.2.0/include/stdarg.h
./include/linux/kernel.h:6,
/openwrt/staging_dir/target-x86_64_musl/usr/include/mac80211-backport/linux/kernel.h:3,
./include/asm-generic/bug.h:20,
/openwrt/staging_dir/target-x86_64_musl/usr/include/mac80211-backport/asm-generic/bug.h:3,
./arch/x86/include/asm/bug.h:93,
./include/linux/bug.h:5,
/openwrt/staging_dir/target-x86_64_musl/usr/include/mac80211-backport/linux/bug.h:3,
/openwrt/staging_dir/target-x86_64_musl/usr/include/mac80211/linux/average.h:5,
/openwrt/build_dir/target-x86_64_musl/linux-x86_64/batman-adv-2021.1/compat-hacks.h:55,
<command-line>:
```

Looks similar to the other path but it actually isn't the same

So in my opinion, @nbd168's patch https://github.com/openwrt/openwrt/commit/9ac47ee46918c45b91f4e4d1fa76b1e26b9d57fe should be (more or less) be cherry picked for batman-adv. Do you want to prepare this as part of this PR? And we need the same fix also for OpenWrt-21.02

Independent of how we continue with this PR, you should definitely get a big thumbs up for debugging this and reporting it. Great work.

batman-adv/src/openwrt-include/stdarg.h
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
```
ecsv, line 1: This file should not be needed when using ```$(KERNEL_NOSTDINC_FLAGS) \```

batman-adv/Makefile
```diff
@@ -70,6 +71,7 @@ NOSTDINC_FLAGS = \
 	-include backport/autoconf.h \
 	-include backport/backport.h \
 	-include $(PKG_BUILD_DIR)/compat-hacks.h \
+	-I$(PKG_BUILD_DIR)/openwrt-include/ \
```
ecsv, line 74: This extra include path is not needed when using ```$(KERNEL_NOSTDINC_FLAGS) \```

djselbeck:
You are right. Your proposed solution is much easier and nicer. I updated the CR. Thanks for the nice words, I had to get my Wifi up again :).

PR #760, opened 2021-11-30
pimbd: add config load on start

uipoui:
Generate config file from uci and set as start config on start of pimbd.

Signed-off-by: Christian Strebe <uipo@freenet.de>

Maintainer: Pierre Pfister
Compile tested: x86_64, NF533, 21.02.1
Run tested: x86_64, NF533, 21.02.0-rc3

Description:
Enable the pimbd to load the configuration on start.

### DIFF
--- a/pimbd/files/pimb.config
+++ b/pimbd/files/pimb.config
@@ -1,0 +1,34 @@
+# The interface section allows to enable pimb on an interface.
+# Several options can be enabled for the interface.
+# All options are disabled by default.
+#
+#config interface '<interface section name>'
+# option pim '<on | off>'     - enable or disable pim
+# option ssbidir '<on | off>' - enable or disable ssbidir
+# option mld '<on | off>'     - enable or disable mld
+# option igmp '<on | off>'    - enable or disable igmp
+# option dev '<device name>'  - set the device name
+
+# The rpa section allows to configure a rendevous point.
+# The interface and the networks must be configured for every rendevous.
+#
+#config rpa '<rendevous point name>'
+# option address '<ip address>'       - rendevous point address
+# list prefix '<net>/<prefix length>' - prefix for the rpa
+
+
+
+# Uncomment the following following section to enable pimb on the interface eth1.
+#config interface 'ETH1'
+#	option pim 'on'
+#	option ssbidir 'on'
+#	option mld 'on'
+#	option igmp 'on'
+#	option dev 'eth1'
+
+
+# Uncomment the following section to configure a rendevous point.
+#config rpa '172_16_0_1'
+#	option address '172.16.0.1'
+#	list prefix '224.0.0.0/4'
+
--- a/pimbd/files/pimbd.init
+++ b/pimbd/files/pimbd.init
@@ -4,18 +4,147 @@ START=90
 STOP=10
 USE_PROCD=1
 
+TYPE=''
+
+DEV=''
+PIM='off'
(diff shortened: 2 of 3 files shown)


PolynomialDivision:
@cotequeiroz Could you review?

pimbd/files/pimb.config
```diff
@@ -0,0 +1,23 @@
...
+
+#config interface 'ETH1'
+#	option pim 'on'
+#	option ssbidir 'on'
+#	option mld 'on'
+#	option igmp 'on'
+#	option dev 'eth1'
+
+
+# Uncomment the following section to configure a rendevous point.
```
cotequeiroz, line 16: This may mislead someone completely unaware of openwrt config syntax.

pimbd/files/pimbd.init
```diff
@@ -4,18 +4,148 @@ START=90
...
+                ;;
+        esac
+    }
     config_load pimb
 
     procd_open_instance
     procd_set_param command /usr/sbin/pimbd
     procd_append_param command -S
     procd_append_param command -L 6
+    procd_append_param command -c /tmp/pimbd.conf
```
cotequeiroz, line 146: Shouldn't this be used only if the file is actually written?  I'm assuming that having a file name here means that the default config file will not be loaded.  If this is indeed true, it would break existing config.

cotequeiroz (commented):
I've never used pimbd, so I'm just reviewing the init script change.

My main question is what happens if the user upgrades the package?  I'm assuming there's a config file already, and that the upgrade will cause it not to be loaded, even if there's no uci configuration.

pimbd/files/pimbd.init
```diff
@@ -4,18 +4,148 @@ START=90
...
+
+    rm -f /tmp/pimbd.conf
+
+    config_cb() {
+        local name="$1"
+        local value="$2"
+        # commands to be run for every option
+        echo "config_cb $name $value"
+        case $TYPE in
+            interface)
```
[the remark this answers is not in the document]
uipoui: I have changed it here and in the following case statement.

pimbd/files/pimbd.init
uipoui: Without this patch the pimbd does not load any configuration file at all. The pimbd just starts and doesn't do anything. It must be configured by using one of the client tools like pimbc. This has to be done after every reboot of the device as well as after every restart of the daemon. There seems to be no configuration at all. I haven't found one in the pimbd.c file.

The file /tmp/pimbd.conf gets only written if there is a configuration in /etc/config/pimb. If the file dosn't exist the pimbd starts and runs as it would do without the patch. This should ensure backward compatibility if an user has some custom code in place.

If there is a configuration in /etc/config/pimb the file /tmp/pimbd.conf will be written and the pimbd will load the configuration.

pimbd/files/pimb.config
```diff
@@ -0,0 +1,23 @@
+# Uncomment the following following section to enable pimb on the interface.
```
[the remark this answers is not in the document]
uipoui: I have updated the description of the configuration.

pimbd/files/pimb.config
uipoui: I have updated the description of the configuration.

uipoui:
I have squashed the commits.

PolynomialDivision:
thanks :)

PR #771, opened 2022-01-14
CI: fix runtime testing for non master branch

PolynomialDivision:
The runtime testing always ran on master branch aka snapshots since the
branch wasn't passed over to the container execution!


aparcar:
Thank you!

PR #772, opened 2022-01-14
olsrd: only regenerate the config on reload

PolynomialDivision:
A reload kills the olsrd daemon. However, the option AllowNoInt allows
to dynamically add and remove interfaces without a restart. Stop
restarting the complete daemon. A restart destroys all the metrics and
will harm the smoothness of the mesh network.

### DIFF
--- a/olsrd/files/olsrd4.init
+++ b/olsrd/files/olsrd4.init
@@ -36,6 +36,12 @@ boot()
 	rc_procd start_service
 }
 
+reload_service()
+{
+	olsrd_generate_config $OLSRD
+	killall -s SIGHUP $OLSRD
+}
+
 start_service() {
 	olsrd_generate_config $OLSRD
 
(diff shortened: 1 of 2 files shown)


PolynomialDivision:
@fhuberts , @HRogge Is it somehow possible to add or remove an interface, without stopping the olsrd daemon? SIGHUP triggers a reconfigure, but this restarts olsrd completely.

PolynomialDivision:
I think about adding some IPC to call this function:
https://github.com/OLSR/olsrd/blob/master/src/interfaces.c#L274

HRogge:
I think I remember someone doing this before and creating a memory leak... because OLSRv1 config handling was never build to be called multiple times. So this might be a bit more work than just calling this function again.

PolynomialDivision:
> I think I remember someone doing this before and creating a memory leak... because OLSRv1 config handling was never build to be called multiple times. So this might be a bit more work than just calling this function again.

I don't call  `config-handling` multiple times. I have that now running several days. Let's see. ;)

PolynomialDivision:
Replaced by: https://github.com/openwrt/routing/pull/774

PR #776, opened 2022-01-31
babeld: add add_interface function

PolynomialDivision:
An interface can be added dynmiacally to babeld by starting babeld with
the local management interface (-G) and saying:

    interface eth0

Add the ubus equivalent of this function:

    ubus call babeld add_interface '{"ifname":"eth0"}'


PolynomialDivision:
@jech could I do something like this? What do you think?

PR #786, opened 2022-02-19
batman packages: Allow parallel builds

ecsv:
Maintainer: @simonwunderlich, @elektra42 
Compile tested: master branch, ath79
Run tested: -

The build systems of these packages are written in a way that it is safe to run the make steps in parallel. The build time can be reduced slightly on modern systems.


elektra42 (approved):
Thanks!

elektra42 (approved):
Thanks again!

elektra42 (approved):

BKPepe:
This seems to be a good candidate to cherry-pick it to OpenWrt 21.02. What do you think @PolynomialDivision ?

PR #791, opened 2022-02-25
bmx6: update to 65cb0d542f16a4b4689f5ad2542c9f24215a6616

BKPepe:
Maintainer: N/A
Compile tested: Turris Omnia, mvebu/cortex-a9, OpenWrt master
Run tested: N/A

**Description:**
Makefile polishing:
- Fixed SPDX License Identifier
- Use AUTORELEASE in PKG_RELEASE
This helps in cases when the PKG_RELEASE is forgotten

- Remove no longer used things like PKG_REV, etc.
Downloaded tarball is now ~80 kB smaller as we use .tar.xz
instead of .tar.gz

### DIFF
--- a/bmx6/Makefile
+++ b/bmx6/Makefile
@@ -25,21 +25,17 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=bmx6
+PKG_VERSION:=0.1-alpha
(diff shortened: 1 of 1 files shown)


mwarning:
looks good

PR #798, opened 2022-02-26
opennds: Makefile refactoring

BKPepe:
Maintainer: @bluewavenet
Compile & run tested:  N/A just moving rows

Description:
- Switch to AUTORELEASE
- Reorder things
- Fixed SPDX License Identifier
- Added PKG_LICENSE_FILES
- Use two spaces instead of tab

### DIFF
--- a/opennds/Makefile
+++ b/opennds/Makefile
@@ -6,40 +6,42 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=opennds
-PKG_FIXUP:=autoreconf
 PKG_VERSION:=9.6.0
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/opennds/opennds/tar.gz/v$(PKG_VERSION)?
-PKG_SOURCE:=opennds-$(PKG_VERSION).tar.gz
 PKG_HASH:=90613e636c668a16eb9d6abfe19715f87cea7000383a53ad6719dec80ff966db
 PKG_BUILD_DIR:=$(BUILD_DIR)/openNDS-$(PKG_VERSION)
 
 PKG_MAINTAINER:=Rob White <rob@blue-wave.net>
+PKG_LICENSE:=GPL-2.0-or-later
+PKG_LICENSE_FILES:=COPYING
+
+PKG_FIXUP:=autoreconf
 PKG_BUILD_PARALLEL:=1
-PKG_LICENSE:=GPL-2.0+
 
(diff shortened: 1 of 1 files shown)


bluewavenet:
@BKPepe 

> You should rather use semantic versioning for that.

Except it is the same source code and same semantic version for OpenWrt master, OpenWrt 21.02 and for generic Linux.
The only difference between master and 21.02 is the dependencies.

But yes, bumping to -3 will be fine.

I will make this new makefile format the template for new releases (and use AUTORELEASE where possible). Thanks!

> Yes, this should hopefully allows to merge this to packages feed.

Has that actually been decided now?
I am sure there are reasons for doing this but it seems to me at first glance it would be better to go the other way and split up packages into logical groups. But then is it more efficient merging all into packages feed?
Just wondering....

bluewavenet:
@BKPepe 
All checks passed. Merge when you are ready.

BKPepe:
> The only difference between master and 21.02 is the dependencies.

Then you can use variants or have packages with suffixes ``-nftables``/``-iptables``because users can decide if they want to use fw3/fw4. From my point of view, switch to nftables was done too early and many packages are not prepared for that.

https://github.com/openwrt/packages/issues/16818

> Has that actually been decided now?

I don't think so. There was some idea to do it 3 years ago and nobody was too proactive to finish it. I just noticed that multiple packages here in the stable branches do not have ``PKG_MIRROR_HASH``, which creates tarball every time and because of that, the builds are taking longer time than they should do.  

Still having Makefiles in good shape is something what is necessary. Even for me, it makes sense to have routing packages in separate repository, but the thing is that this repository does not have many members (including me), who can merge stuff, etc.

bluewavenet:
@BKPepe 

> Then you can use variants or have packages with suffixes -nftables/-iptables

Suffix on the 21.02 branch might be a way forward I think.

openNDS is completely independent of fw3/fw4. The real issue here is that iptables-nft is not available in 21.02. If it was, the problem would go away.

openNDS extensively uses iptables and ipset and works perfectly with iptables-nft.

It is going to be a very edge case if a package cannot use iptables-nft, given the success we have with openNDS using it.
Do you have any influence getting iptables-nft into 21.02?

PolynomialDivision:
Should I merge?

bluewavenet:
@BKPepe 

> I don't have merge permissions here, sorry. :-(

No rush, but perhaps @PolynomialDivision or @mwarning  can oblige?

BKPepe:
> Do you have any influence getting iptables-nft into 21.02?

I will take a look, that's a good idea.

bluewavenet:
@PolynomialDivision 
Yes please :-D

PolynomialDivision:
@BKPepe I am happy if someone gives you merge permission cc @aparcar

Issue #731, opened 2021-10-06
Certain upstream switch to `firewall4` aka `nftables` instead of `iptables`


aparcar:
Hi all, especially @openwrt/routing-write,

for the next OpenWrt release `firewall4` is [considered](https://github.com/openwrt/openwrt/pull/4642) as a replacement of the current `iptables` based `firewall` package. While the configuration stays within `/etc/config/firewall`, packages using `iptables` directly may see trouble.

This is a heads up for everyone maintaining such packages but also please post packages here that would be affected so a smother migration is possible.

Compatible with `firewall4`:
- [x] OLSR

Heads up for packages.git: https://github.com/openwrt/packages/issues/16818
Heads up for luci.git: https://github.com/openwrt/luci/issues/5409

PolynomialDivision:
OLSR is affected.

pmelange:
> OLSR is affected.

with a ```git grep -i iptables``` i see that nodogsplash and opennds both require "iptables-mod-ipopt".  What would be the equivalent package in fw4?

In OLSR, it seems like only olsrd/files/olsrd.sh is using iptables and ip6tables in a script.  Is there a syntax change on the command line going from iptables to fw4?

I didn't look at any of the source code of any of the packages, just at the repo itself.

zioproto:
@PolynomialDivision is it olsrd core code affected or it is an extension/plugin ? Is it just the [Smart Gateway extension](https://github.com/OLSR/olsrd/blob/v0.9.8/README-Olsr-Extensions#L221) broken ?

aparcar:
@openwrt/routing-write  heads up, firewall4 gets into a fine state (thanks to @stintel and @jow- )and I'm planing to make it the default within the next week. For all scripts that require _legacy_ `iptables` please test your apps with `iptables-nft` and report back!

bluewavenet:
@aparcar 
Where is `iptables-nft`? (It is probably hidden in clear sight, ie I can't find it :-D )
openNDS works fine with it on "Generic Linux" eg Debian, et al, so should be good here as well at least as an interim solution.

wfleurant:
i found something in LuCI. When I Save & Apply a rule under "Traffic Rules" it does not restart the firewall.  
when manually running `/etc/init.d/firewall restart` then all of the rules I enabled/disabled work as expected.  
compiled against your aparcar/openwrt:fw4 (8af05a1) .. tested on rpi4

castillofrancodamian:
[...]

PR #812, opened 2022-02-27
olsrd: switch to firewall4

PolynomialDivision:
Firewall4 uses nftables instead of iptables. Use iptables-nft for installing smart-gw-rules.

Fixes: #731 ("Certain upstream switch to firewall4 aka nftables instead of iptables")

### DIFF
--- a/olsrd/Makefile
+++ b/olsrd/Makefile
@@ -34,7 +34,7 @@ endef
 define Package/olsrd
   $(call Package/olsrd/template)
   MENU:=1
-  DEPENDS:=+libpthread +libubus
+  DEPENDS:=+libpthread +libubus +iptables +ip6tables
 endef
 
 define Package/olsrd/conffiles


PolynomialDivision:
I don't use smart gateway. Can anyone test?

olsrd/files/olsrd.sh
```diff
@@ -746,8 +746,8 @@ olsrd_setup_smartgw_rules() {
 		nowan=0
 	fi
 
-	IP4T="$( command -v iptables )"
-	IP6T="$( command -v ip6tables )"
+	IP4T="$( command -v iptables-nft )"
```
BKPepe, line 749: This change does not seems to be necessary.


```
root@turris:~# iptables
iptables v1.8.7 (nf_tables): no command specified
Try `iptables -h' or 'iptables --help' for more information.

```

olsrd/Makefile
```diff
@@ -34,7 +34,7 @@ endef
 define Package/olsrd
   $(call Package/olsrd/template)
   MENU:=1
-  DEPENDS:=+libpthread +libubus
+  DEPENDS:=+libpthread +libubus +iptables-nft +ip6tables-nft
```
BKPepe, line 37: This breaks compatibility with fw3.  By quick looking at commit https://github.com/openwrt/openwrt/commit/795e7155cbe3e78669f6821bf7aecb7c4e1e1afb it should be enough to add iptables and iptables6 w/o any suffix. :)

BKPepe, line 37: This breaks compatibility with fw3.  By quick looking at commit https://github.com/openwrt/openwrt/commit/795e7155cbe3e78669f6821bf7aecb7c4e1e1afb it should be enough to add iptables and iptables6 w/o any suffix. :)

BKPepe (changes requested):

PolynomialDivision:
Thanks! Changed it.

BKPepe:
Commit subject and commit description should be changed.

PolynomialDivision:
Did that.

BKPepe:
What about this? I wouldn't say I like copy&pasting commit messages from different commits and also the references always begin with 1 instead of 0.

```
The upcoming OpenWrt version (currently in the master branch) ships firewall4 by default [1], which uses nftables instead of iptables. 

Users might choose to use a compatible layer [2] with iptables-nft, decide which iptables they want to use, and make appropriate changes related to fw3/f4[3].

This package requires to use iptables (no matter which variant) for smart-gw-rules 

[1] https://github.com/openwrt/openwrt/commit/08d9f6e3020b4a149b2007b6ed7d684c49af9bbf

[2] https://www.redhat.com/en/blog/using-iptables-nft-hybrid-linux-firewall

[3] https://github.com/openwrt/openwrt/commit/795e7155cbe3e78669f6821bf7aecb7c4e1e1afb
```

Feel free to rewrite it based on your thoughts.

From my point of view, you should not use any **[Fixes:](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests)** or any other keyword in the commit message since when you have it there and you will do force-push, you will only spam issue with mentions. You should rather put it to the pull request.

PolynomialDivision:
Sounds awesome! :)

BKPepe (approved):

PR #815, opened 2022-03-05
[21.02] cjdns: fix uci-defaults (#714)

PolynomialDivision:
Since https://git.openwrt.org/?p=openwrt/openwrt.git;a=commitdiff;h=4b9a67362d70c544b85078b8d5c661f43f7472d9,
uci network config interface sections use 'device' instead of 'ifname'

Even if 'device' seems to always be set by
package/base-files/files/bin/config_generate, we keep the fallbacks
in case someone install the package via opkg and has a non migrated
network config

Signed-off-by: Etienne Champetier <champetier.etienne@gmail.com>
(cherry picked from commit 383c52ff8d9c0577a262d014aca66bd4095f4e38)

### DIFF
--- a/cjdns/Makefile
+++ b/cjdns/Makefile
@@ -18,7 +18,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=cjdns
 PKG_VERSION:=v21
-PKG_RELEASE:=1
+PKG_RELEASE:=2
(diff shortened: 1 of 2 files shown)


PolynomialDivision:
I will mege since this is only a cherry-pick.

PR #816, opened 2022-03-05
opennds: don't use iptables-nft as dependency

PolynomialDivision:
Since, commit 795e7155cbe3 ("iptables: rename to ip(6)tables-legacy, add PROVIDES") it is enough to include iptables as dependency to install iptables-nft.

@BKPepe

### DIFF
--- a/opennds/Makefile
+++ b/opennds/Makefile
@@ -7,7 +7,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=opennds
 PKG_VERSION:=9.6.0
-PKG_RELEASE:=3
+PKG_RELEASE:=4
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/opennds/opennds/tar.gz/v$(PKG_VERSION)?
@@ -27,7 +27,7 @@ define Package/opennds
   SUBMENU:=Captive Portals
   SECTION:=net
   CATEGORY:=Network
-  DEPENDS:=+iptables-nft +kmod-ipt-conntrack +kmod-ipt-nat +libmicrohttpd-no-ssl
+  DEPENDS:=+iptables +kmod-ipt-conntrack +kmod-ipt-nat +libmicrohttpd-no-ssl
   TITLE:=Open public network gateway daemon
   URL:=https://github.com/opennds/opennds
   CONFLICTS:=nodogsplash nodogsplash2
@@ -41,7 +41,7 @@ define Package/opennds/description
   The package incorporates the FAS API allowing many flexible customisation options.
   The creation of sophisticated third party authentication applications is fully supported.
   Internet hosted https portals can be implemented with no security errors, to inspire maximum user confidence.
-  This version requires iptables-nft.
+  This version requires iptables.
 endef
 
 define Package/opennds/install


BKPepe:
Commit description could be improved. There should be at least referenced the commit to find it easier. But it should be tested with ``iptables-legacy`` and ``iptables-nft``. 

I still don't switch to fw4 completely, though.

PolynomialDivision:
@bluewavenet

bluewavenet:
@PolynomialDivision @BKPepe 
This sounds ideal and solves a lot of problems. Any package using iptables is likely to be perfectly ok without any changes.
I'll do some testing as soon as I can.

But the wording of the iptables commit is very ambiguous.

Can you confirm what it actually means is that - in master:
* iptables-nft is renamed as iptables
* The previously missing iptables package is now available as iptables-legacy

PolynomialDivision:
Yes exactly, and they both **provide** iptables. User needs to decide what to select.

bluewavenet:
@PolynomialDivision @BKPepe 
Ok I've done some testing.
Now I get iptables-legacy.
**But what if I want iptables-nft by default?** (I think I do for opennds - iptables calls seem to be significantly faster using iptables-nft - no definitive testing but startup of opennds is definitely quicker with the nft version)

In my opinion this is the wrong way around, **we should be defaulting to iptables-nft if iptables is installed (post 21.02)**
Now we can end up by default with iptables-legacy and fw4, probably not a problem but inconsistent and not very clean.

_**Defaulting to iptables-nft if the iptables package is installed is very unlikely to cause problems for any other package**_ (given the intensive use in openNDS including optional ipset use - if openNDS works, anything will), and if a problem was found in a particular case then the legacy version can be specified.

bluewavenet:
@PolynomialDivision @BKPepe 
If we install iptables, we get iptables-legacy.
Then if I install  iptables-nft, **it gets installed as well as iptables-legacy.** 

I would expect a PROVIDES error and the install of iptables-nft to fail....

PolynomialDivision:
I believe that iptables-nft will be default in firewall4. However, if a user wants to default to firewall3 without nftables, when this user has to also install nftables. So I would go for `iptables`. :)
However, at the end its your decision. :)

bluewavenet:
@PolynomialDivision 
It is not really to to with fw4/fw3 choices.
Post 21.02.xx we have by default, nftables with no iptables support installed.

If iptables is required (post 21.02.xx), then surely installing the iptables package should automatically default to iptables(nftables).
**Currently** we have iptables(legacy) being the default. _**This is the wrong way round.**_

If someone chooses to downgrade from fw4 to fw3 and remove nftables, then yes then, **and only then,** should they also install iptables(legacy).

_I have not tried it, but would fw3 not run perfectly well with iptables(nftables)?_

> Yes exactly, and they both provide iptables.

**There is an issue here** as it is possible to install **both** iptables(nftables) and iptables(legacy)

bluewavenet:
@PolynomialDivision 
I just reread your initial comment:

> it is enough to include iptables as dependency to install iptables-nft.

_That is my point, include iptables as a dependency and you get iptables(legacy)......_

PolynomialDivision:
Okay, I will close. :)

bluewavenet:
@PolynomialDivision 
Thanks. There is another new release coming soon anyway.

However, I feel that everyone has  missed the point I am trying to make:
**Post 21.02.x `opkg install iptables` _should_ result in iptables-nft being installed**

Your original comment states:

> it is enough to include iptables as dependency to install **iptables-nft.**

**However in practice this is _NOT_ the case. Instead we get _iptables-legacy_**
(or did last time I flashed a fresh snapshot)

PR #825, opened 2022-03-22
mesh11sd: [New Package] Release v1.0.0

bluewavenet:
Compile tested: All
Run tested: arm_cortex-a7_neon-vfpv4, mipsel_24kc, x86-64, on 21.02.2 and snapshot.

Description:
Mesh11sd is a dynamic parameter configuration daemon for 802.11s mesh networks.
It was originally designed to leverage 802.11s mesh networking at Captive Portal venues.
This is the open source version and it enables easy and automated mesh network operation with multiple mesh nodes.
It allows all mesh parameters supported by the wireless driver to be set in the uci config file.
Settings take effect immediately without having to restart the wireless network.
Default settings give rapid and reliable layer 2 mesh convergence.
Without mesh11sd, many mesh parameters cannot be set in the uci wireless config file as the mesh interface must be up before the parameters can be set.
Some of those that are supported, would fail to be implemented when the network is (re)started resulting in errors or dropped nodes.
The mesh11sd daemon dynamically checks configured parameters and sets them as required.
This version does not require a Captive Portal to be running.

### DIFF
--- a/mesh11sd/Makefile
+++ b/mesh11sd/Makefile
@@ -1,0 +1,63 @@
+# SPDX-License-Identifier: GPL-2.0-only
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+# Copyright (C) 2022 BlueWave Projects and Services  <licence@blue-wave.net>
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=mesh11sd
+PKG_VERSION:=1.0.0
+PKG_RELEASE:=$(AUTORELEASE)
+
+PKG_MAINTAINER:=Rob White <rob@blue-wave.net>
+PKG_LICENSE:=GPL-2.0-or-later
+PKG_LICENSE_FILES:=LICENSE
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/opennds/mesh11sd/tar.gz/v$(PKG_VERSION)?
+PKG_HASH:=a9f2ac4e1cdf5c4fe933ad68adf7c62d6d4a5e20a7f61f79dfa26a309f97049e
+PKG_BUILD_DIR:=$(BUILD_DIR)/mesh11sd-$(PKG_VERSION)
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/mesh11sd
+  SUBMENU:=Captive Portals
+  SECTION:=net
+  CATEGORY:=Network
+  TITLE:=Dynamic 802.11s Mesh Configuration Daemon
+  PKGARCH:=all
+  URL:=https://github.com/opennds/mesh11sd
+endef
+
+define Package/mesh11sd/description
+  Mesh11sd is a dynamic parameter configuration daemon for 802.11s mesh networks.
+  It was originally designed to leverage 802.11s mesh networking at Captive Portal venues.
+  This is the open source version and it enables easy and automated mesh network operation with multiple mesh nodes.
+  It allows all mesh parameters supported by the wireless driver to be set in the uci config file.
+  Settings take effect immediately without having to restart the wireless network.
+  Default settings give rapid and reliable layer 2 mesh convergence.
+  Without mesh11sd, many mesh parameters cannot be set in the uci wireless config file as the mesh interface must be up before the parameters can be set.
+  Some of those that are supported, would fail to be implemented when the network is (re)started resulting in errors or dropped nodes.
+  The mesh11sd daemon dynamically checks configured parameters and sets them as required.
+  This version does not require a Captive Portal to be running.
+endef
+
+define Package/mesh11sd/install
+	$(INSTALL_DIR) $(1)/usr/sbin
+	$(INSTALL_DIR) $(1)/etc/config
+	$(INSTALL_DIR) $(1)/etc/init.d
+	$(CP) $(PKG_BUILD_DIR)/src/mesh11sd $(1)/usr/sbin
+	$(CP) $(PKG_BUILD_DIR)/linux_openwrt/mesh11sd/files/etc/config/mesh11sd $(1)/etc/config/
+	$(CP) $(PKG_BUILD_DIR)/linux_openwrt/mesh11sd/files/etc/init.d/mesh11sd $(1)/etc/init.d/
+endef
+
+define Package/mesh11sd/conffiles
+/etc/config/mesh11sd
+endef
+
+define Build/Compile
+endef
+
+$(eval $(call BuildPackage,mesh11sd))


bluewavenet:
@PolynomialDivision @mwarning 
I think I've done this correctly - it is script based so just packaging, no compiling.

mwarning:
I feel a bit blind here. Isn't mesh11sd/files/etc/init.d/mesh11sd etc. missing?

bluewavenet:
@mwarning 
It's all here:
https://github.com/openNDS/mesh11sd/releases/tag/v1.0.0

bluewavenet:
@mwarning @PolynomialDivision 
This builds fine with the SDK so should be good to merge here unless anyone can see a problem... I used the example of Naywatch for the Makefile with recipe entries to install the files from the release tar from Github.

**A summary of what mesh11sd does:**
It allows any wireless driver supported mesh parameter to be set from a uci config file.
Without mesh11sd, many mesh parameters fail to be set from /etc/config/wireless as the mesh interface must be fully up before the parameters are activated in the driver, and of course the contents of /etc/config/wireless are used only on startup (or restart).

Mesh11sd allows any mesh parameter to be set dynamically and tuned "on the fly" without a wireless restart opening many options that can be implemented, many of which will be explored in later versions.

Mesh11sd also has a cli interface that allows monitoring of the mesh. This will also be expanded in later versions.

Here is an example output of `mesh11sd status` from a node in an active mesh network:
```
{
   "setup":{
     "version":"1.0.0",
     "enabled":"1",
     "checkinterval":"15",
     "interface_timeout":"10",
     "debuglevel":"1"
}
  "interfaces":{
    "mesh0":{
      "mesh_retry_timeout":"100",
      "mesh_confirm_timeout":"100",
      "mesh_holding_timeout":"100",
      "mesh_max_peer_links":"150",
      "mesh_max_retries":"3",
      "mesh_ttl":"31",
      "mesh_element_ttl":"31",
      "mesh_auto_open_plinks":"0",
      "mesh_hwmp_max_preq_retries":"4",
      "mesh_path_refresh_time":"1000",
      "mesh_min_discovery_timeout":"100",
      "mesh_hwmp_active_path_timeout":"5000",
      "mesh_hwmp_preq_min_interval":"10",
      "mesh_hwmp_net_diameter_traversal_time":"50",
      "mesh_hwmp_rootmode":"3",
      "mesh_hwmp_rann_interval":"5000",
      "mesh_gate_announcements":"1",
      "mesh_fwding":"1",
      "mesh_sync_offset_max_neighor":"50",
      "mesh_rssi_threshold":"-80",
      "mesh_hwmp_active_path_to_root_timeout":"6000",
      "mesh_hwmp_root_interval":"5000",
      "mesh_hwmp_confirmation_interval":"2000",
      "mesh_power_mode":"active",
      "mesh_awake_window":"10",
      "mesh_plink_timeout":"0",
      "mesh_connected_to_gate":"0",
      "mesh_nolearn":"0",
      "mesh_connected_to_as":"0",
      "mesh_id":"PublicFreeMesh",
      "device":"radio0",
      "channel":"11",
      "active_peers":"17"
    }
  }
}

```

mwarning:
looks good , thank you :)

mesh11sd/Makefile
```diff
@@ -0,0 +1,63 @@
...
+PKG_RELEASE:=$(AUTORELEASE)
+
+PKG_MAINTAINER:=Rob White <rob@blue-wave.net>
+PKG_LICENSE:=GPL-2.0-or-later
+PKG_LICENSE_FILES:=LICENSE
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_SOURCE_URL:=https://codeload.github.com/opennds/mesh11sd/tar.gz/v$(PKG_VERSION)?
+PKG_HASH:=a9f2ac4e1cdf5c4fe933ad68adf7c62d6d4a5e20a7f61f79dfa26a309f97049e
+PKG_BUILD_DIR:=$(BUILD_DIR)/mesh11sd-$(PKG_VERSION)
```
BKPepe, line 21: This could be removed, right? It is already default.

mesh11sd/Makefile
```diff
@@ -0,0 +1,63 @@
...
+  Some of those that are supported, would fail to be implemented when the network is (re)started resulting in errors or dropped nodes.
+  The mesh11sd daemon dynamically checks configured parameters and sets them as required.
+  This version does not require a Captive Portal to be running.
+endef
+
+define Package/mesh11sd/install
+	$(INSTALL_DIR) $(1)/usr/sbin
+	$(INSTALL_DIR) $(1)/etc/config
+	$(INSTALL_DIR) $(1)/etc/init.d
+	$(CP) $(PKG_BUILD_DIR)/src/mesh11sd $(1)/usr/sbin
```
BKPepe, line 51: Shouldn't you use INSTALL_BIN?

mesh11sd/Makefile
```diff
@@ -0,0 +1,63 @@
...
+	$(CP) $(PKG_BUILD_DIR)/linux_openwrt/mesh11sd/files/etc/config/mesh11sd $(1)/etc/config/
```
BKPepe, line 52: Shouldn't you use INSTALL_CONF?

mesh11sd/Makefile
bluewavenet: You mean PKG_SOURCE is default? This line is included after taking other package makefiles by example, as it the whole of this makefile.

This makefile clearly works with no problems, but I am always happy to receive guidance.

The problem I have is that as far as I can see there is no documentation on package makefiles - do you have a link to something I can read?

mesh11sd/Makefile
bluewavenet: Should I? What is the difference between INSTALL_BIN and INSTALL_DIR?
This package installs plain text files to various locations. The "BIN" part implies "binary", but that would be guesswork on my behalf.

mesh11sd/Makefile
BKPepe:
```suggestion
```

What are you doing is:
``PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)-$(PKG_VERSION)``
and that's already default.

There is documentation, but I am afraid that it could be outdated.

mesh11sd/Makefile
bluewavenet: See my comment on INSTALL_BIN.

mesh11sd/Makefile
BKPepe: https://openwrt.org/docs/guide-developer/packages#file_installation_macros

bluewavenet:
@BKPepe 
Thank you for your input, it is much appreciated.

mesh11sd/Makefile
bluewavenet: Thank you. Not sure why I could not find it.

mesh11sd/Makefile
bluewavenet: That makes sense...

bluewavenet:
@BKPepe 
There is a new release of mesh11sd upcoming, so I have revisited INSTALL_DIR, INSTALL_DIR, INSTALL_DIR, CP macros.
After trial an error, my understanding is as follows:
 * INSTALL_DIR - creates the directory with mode 0755
 * INSTALL_BIN - copies file and sets mode to 0755
 * INSTALL_CONF - copies file and sets mode to 0600
 * CP - copies file(s), keeping the mode of the source, using `cp -fpR [source] [destination]`

So in conclusion, I think I should have:
```
define Package/mesh11sd/install
	$(INSTALL_DIR) $(1)/usr/sbin
	$(INSTALL_DIR) $(1)/etc/config
	$(INSTALL_BIN) $(PKG_BUILD_DIR)/src/mesh11sd $(1)/usr/sbin
	$(INSTALL_CONF) $(PKG_BUILD_DIR)/linux_openwrt/mesh11sd/files/etc/config/mesh11sd $(1)/etc/config/
	$(INSTALL_BIN) $(PKG_BUILD_DIR)/linux_openwrt/mesh11sd/files/etc/init.d/mesh11sd $(1)/etc/init.d/
endef

```

This has been tested using the SDK, and seems to work fine.

If you agree, I will update https://openwrt.org/docs/guide-developer/packages#file_installation_macros accordingly.

Re. Removing the line:
`PKG_BUILD_DIR:=$(BUILD_DIR)/mesh11sd-$(PKG_VERSION)`

Using the SDK, this does not seem to work, at least for me, and the build fails. So I would rather keep this in place unless it has a particular negative effect that you can see.

PR #833, opened 2022-03-28
babeld: add add_filter function

PolynomialDivision:
I just tried to add simple filter. However, this does not work. Any idea what is missing to add a filter @jech?

E.g.

     in if eth0 metric 128

should be

    ubus call babeld add_filter '{"ifname":"eth0", "type":"input", "metric":128}'

### DIFF
--- a/babeld/patches/000-export-add-filters-and-simplify-interface.patch
+++ b/babeld/patches/000-export-add-filters-and-simplify-interface.patch
@@ -1,0 +1,116 @@
+From 7c053fe7584b7b4fe4effc09624ae620304d6717 Mon Sep 17 00:00:00 2001
+From: Juliusz Chroboczek <jch@irif.fr>
+Date: Tue, 29 Mar 2022 19:26:50 +0200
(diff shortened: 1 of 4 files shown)


PolynomialDivision:
Maybe I was to tired yesterday. I just tested everything again and it works flawlessly. Are you fine with adding this already @jech? I will go on defining all possible filter functions, however, I urgently need this functionality only for ipv6 and all routes on an interface.

jech:
I'll change the interface a little bit, to make it less error-prone, but the principle will remain the same.  Please give me a couple of days.

PolynomialDivision:
> I'll change the interface a little bit, to make it less error-prone, but the principle will remain the same. Please give me a couple of days.

Works perfectly! Thanks a lot. I would merge?

jech:
Merged.

PolynomialDivision:
Great than I will also merge this PR. :) Maybe time for a new release?

jech:
I was planning to merge v4-via-v6 first, but it might indeed be worthwile to release before then.

PR #846, opened 2022-04-01
opennds: avoid circular dependency

cotequeiroz:
Maintainer: @bluewavenet 
Compile tested: mediatek/mt7622, aarch64-cortexa53
Run tested: none

Description:
Have the package depend on `@IPTABLES_NFTABLES`, which is needed to be able to select `iptables-nft`.  It is also necessary to make that selection dependent on `@IPTABLES_NFTABLES` as well, so that the config generator understands it is a conditional selection.

With this change, the following two lines are added to the generated tmp/.config-package.in.in:
```
	depends on IPTABLES_NFTABLES
	select iptables-nft if IPTABLES_NFTABLES
```
This avoids the following error, whose message does not help much:
```
Collecting package info: done
tmp/.config-package.in:102128:error: recursive dependency detected!
tmp/.config-package.in:102128: 	symbol PACKAGE_iptables-legacy is selected by PACKAGE_iptables-nft
tmp/.config-package.in:102935: 	symbol PACKAGE_iptables-nft is selected by PACKAGE_opennds
tmp/.config-package.in:101158: 	symbol PACKAGE_opennds depends on PACKAGE_iptables-legacy
For a resolution refer to Documentation/kbuild/kconfig-language.rst
subsection "Kconfig recursive dependency limitations"
```
What you need to avoid is the selection of `iptables-nft` when `IPTABLES_NFTABLES` is not selected, since `iptables-nft` depends on `IPTABLES_NFTABLES`.

Signed-off-by: Eneas U de Queiroz <cotequeiroz@gmail.com>

### DIFF
--- a/opennds/Makefile
+++ b/opennds/Makefile
@@ -7,7 +7,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=opennds
 PKG_VERSION:=9.7.0
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/opennds/opennds/tar.gz/v$(PKG_VERSION)?
@@ -27,7 +27,7 @@ define Package/opennds
   SUBMENU:=Captive Portals
   SECTION:=net
   CATEGORY:=Network
-  DEPENDS:=+iptables-nft +kmod-ipt-conntrack +kmod-ipt-nat +libmicrohttpd-no-ssl
+  DEPENDS:=@IPTABLES_NFTABLES +iptables-nft +kmod-ipt-conntrack +kmod-ipt-nat +libmicrohttpd-no-ssl
   TITLE:=Open public network gateway daemon
   URL:=https://github.com/opennds/opennds
   CONFLICTS:=nodogsplash nodogsplash2 iptables-legacy


bluewavenet:
@cotequeiroz 
Thank you for this.

This is, however, a work-around for the underlying problem which is an error in the iptables makefile (duplicate PROVIDES). This has the effect of making iptables-legacy the default (as the iptables-legacy section comes first in the makefile).

It is this line that tells the story:
`tmp/.config-package.in:102128: 	symbol PACKAGE_iptables-legacy is selected by PACKAGE_iptables-nft`

When building, it is impossible to select iptables-nft (without your work-around).

The resulting recursive dependency when building does not prevent build-bots or the SDK from creating the openNDS ipk (I'm not sure what happens if using the tool chain).

An opkg install of the resulting openNDS ipk works perfectly. 
The openNDS **_requirement_** is for the opkg install to fail if iptables-legacy is already installed.

For OpenWrt **_after 21.02.x_**, by default, no form of iptables is installed as OpenWrt has moved to nftables. 
There should **never be any need to install iptables-legacy** apart from an unlikely and as yet unidentified edge case in some other package.
Yet `opkg install iptables` results in iptables-legacy being installed.

>  Run tested: none

As you have not run tested, is there any guarantee that this workaround has no effect on the opkg install?

bluewavenet:
@cotequeiroz 

> To be clear: I'm not touching anything that opkg will do (for better or for worse). I'm just handling build-time dependencies for now.

**Yes, but it is install-time dependencies that matter.**

I do think though that this PR would be unnecessary if the error in the iptables makefile was fixed.
If `opkg install iptables` resulted in iptbles-nft being installed (as it should be but unfortunately does not), then openNDS's dependency could be simply `DEPENDS:=+iptables ...... `.

The iptables makefile:
https://github.com/openwrt/openwrt/blob/master/package/network/utils/iptables/Makefile#L53

The "define" section starting at line 53 comes first and takes precedence. It specifies `PROVIDES:=iptables`. **This makes iptables-legacy the default and is incorrect.**

Then the "define section starting at line 132 also specifies `PROVIDES:=iptables`. **This is the one that should be the default**.

Either line 57 should be deleted or the two define blocks should be swapped over.

As you are building from scratch, could you try deleting line 53 from the iptables makefile to see if the problem is fixed?
(I would try it myself but I do not currently have a machine suitable for the full build and have to restrict myself to using the SDK)

PR #848, opened 2022-04-09
[21.02] mesh11sd: [New Package] Release v1.0.0

bluewavenet:
Compile tested: All
Run tested: arm_cortex-a7_neon-vfpv4, mipsel_24kc, x86-64, on 21.02.2

Description:
  Mesh11sd is a dynamic parameter configuration daemon for 802.11s mesh networks.
  It was originally designed to leverage 802.11s mesh networking at Captive Portal venues.
  This is the open source version and it enables easy and automated mesh network operation with multiple mesh nodes.
  It allows all mesh parameters supported by the wireless driver to be set in the uci config file.
  Settings take effect immediately without having to restart the wireless network.
  Default settings give rapid and reliable layer 2 mesh convergence.
  Without mesh11sd, many mesh parameters cannot be set in the uci wireless config file as the mesh interface must be up before the parameters can be set.
  Some of those that are supported, would fail to be implemented when the network is (re)started resulting in errors or dropped nodes.
  The mesh11sd daemon dynamically checks configured parameters and sets them as required.
  This version does not require a Captive Portal to be running.

(cherry picked from commit b4ab8968946dd3210fb84a812a7f46d0fbc03899)

### DIFF
--- a/mesh11sd/Makefile
+++ b/mesh11sd/Makefile
@@ -1,0 +1,63 @@
+# SPDX-License-Identifier: GPL-2.0-only
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+# Copyright (C) 2022 BlueWave Projects and Services  <licence@blue-wave.net>
+#
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=mesh11sd
+PKG_VERSION:=1.0.0
+PKG_RELEASE:=$(AUTORELEASE)
+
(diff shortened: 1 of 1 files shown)


bluewavenet:
@PolynomialDivision 
Here is another one - cherry picked of course ;-)

PolynomialDivision:
lgtm

bluewavenet:
@PolynomialDivision 
Thanks again :-D

PR #860, opened 2022-05-08
babeld: update to 1.12

PolynomialDivision:
Changelog:

5 May 2022: babeld-1.12

  * Implement v4-via-v6 routing (RFC 9229), which allows a router with
    IPv4 addresses only to route IPv4.  Thanks to Théophile Bastian.
  * Enable extended Netlink acks when available.
    Thanks to Toke Høyland-Jørgensen.
  * Fix restoring of interface configuration to avoid unbounded memory
    consumption.  Thanks to andrew-hoff.
  * Fix handling of deny filters in the install chain.

Refreshed patches:
- 600-add-ubus.patch

### DIFF
--- a/babeld/Makefile
+++ b/babeld/Makefile
@@ -6,12 +6,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=babeld
-PKG_VERSION:=1.11
+PKG_VERSION:=1.12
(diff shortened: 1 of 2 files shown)


PolynomialDivision:
Runtested on `Linksys E8450`. Works perfectly.

PR #876, opened 2022-06-02
ci: add commit linting

PolynomialDivision:


BKPepe:
This should be first added to packages repository and reviewed there.

PolynomialDivision:
> This should be first added to packages repository and reviewed there.

It had some serious limitations and I don't find it useful anymore. Closing.

BKPepe:
For the sake of completeness, would you mind sharing which serious limitations it have?

PolynomialDivision:
> For the sake of completeness, would you mind sharing which serious limitations it have?

I am sorry, but I can not remember correctly. Do you think such linting would be useful in general? Maybe I can go into this again and check if I can tweak all configurations that it is useful for us/me.

PR #877, opened 2022-06-03
olsrd: Robustify default procd respawn_retry

Hurricos:
Should `olrsd` crash *just once* after starting less than 3600s ago,
it will never restart. Fix that by making `olsrd`'s `respawn_retry`
non-zero.

Signed-off-by: <hurricos@gmail.com>

Maintainer: @nhainke
Compile tested: Need I really compile-test a procd option change?
Run tested: 4 x Meraki MR16 @ OpenWrt 21.02.0 -- but with this change made in a `config procd 'procd'` section in `/etc/config/olsrd`, rather than in the init script of the package.

### DIFF
--- a/olsrd/files/olsrd4.init
+++ b/olsrd/files/olsrd4.init
@@ -48,7 +48,7 @@ start_service() {
 
 	config_get _respawn_threshold procd respawn_threshold 3600
 	config_get _respawn_timeout procd respawn_timeout 15
-	config_get _respawn_retry procd respawn_retry 0
+	config_get _respawn_retry procd respawn_retry 5
 
 	procd_set_param command "$BIN"
 	procd_append_param command -f ${CONF}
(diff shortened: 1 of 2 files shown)


Hurricos:
I'm submitting PR because crashes in `olsrd` that subsequently make the nodes unreachable have painful consequences.

Let me know if I should instead be sending this to e.g. a Patchwork feed, or to the mailing list.

Hurricos:
... @PolynomialDivision. Sorry; Github didn't like my initial attempt to indicate who I was trying to ping.

Hurricos:
Withdrawing this. Testing was done on too old of a version of olsrd which did NOT have procd support. In addition, as stated in the original commit message for bc58bd99e81b72f8061473eec5d3454b3504b163, and as tested by me on a newer version of olsrd, the default value of `respawn_retry` of 0 will *always* restart olsrd.

PR #911, opened 2022-10-20
babeld: add del_filters function

PolynomialDivision:
This commit adds the ubus function to dynamically delete filters on runtime:

  ubus call babeld del_filters '{"ifname":"eth0"}'

Maintainer: me, @kerneis 
Compile tested: belkin rt3200
Run tested: belkin rt3200

### DIFF
--- a/babeld/patches/100-Add-functions-to-delete-filters-from-an-interface.patch
+++ b/babeld/patches/100-Add-functions-to-delete-filters-from-an-interface.patch
@@ -1,0 +1,65 @@
+From 7dc87f39dc5d1db7654f6ad89eca77766812affd Mon Sep 17 00:00:00 2001
+From: Nick Hainke <vincent@systemli.org>
+Date: Tue, 18 Oct 2022 21:34:03 +0200
+Subject: [PATCH] Add functions to delete filters from an interface
+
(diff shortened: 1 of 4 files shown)


babeld/patches/100-Add-functions-to-delete-filters-from-an-interface.patch
```diff
@@ -0,0 +1,65 @@
+From 7dc87f39dc5d1db7654f6ad89eca77766812affd Mon Sep 17 00:00:00 2001
+From: Nick Hainke <vincent@systemli.org>
+Date: Tue, 18 Oct 2022 21:34:03 +0200
+Subject: [PATCH] Add functions to delete filters from an interface
```
BKPepe, line 4: Reference: https://github.com/jech/babeld/pull/95

BKPepe:
May I know which is your use case or if you are using this anyhow these days?  Code was reviewed that's ok, but babeld's developer said that adding these two function to upstream does not make any sense as it is not used, but you are using this here. 

So, I think this can go through, though, and it will give time to babeld's developer to come up with something better, as he said.

BKPepe:
Closing due to inactivity.

PR #915, opened 2022-10-29
ci: update github actions to v3

BKPepe:
Update checkout and upload-artifact action to v3 to mute nodejs deprecation warning.

Backported from https://github.com/openwrt/packages/pull/19705

### DIFF
--- a/.github/workflows/formal.yml
+++ b/.github/workflows/formal.yml
@@ -11,7 +11,7 @@ jobs:
       fail-fast: false
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
           ref: ${{ github.event.pull_request.head.sha }}
           fetch-depth: 0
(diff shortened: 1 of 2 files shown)


PolynomialDivision:
Is that really a backport and can you cherry-pick into something that is not in this git?

Until now the commits were just applied without a cherry-pick:
https://github.com/openwrt/routing/commit/e2d72c57756751230d61bd7a738a35dad912e644

BKPepe:
Yes, you can work at multiple Git repositories by using [Git remotes](https://git-scm.com/book/en/v2/Git-Basics-Working-with-Remotes) if the folder structure is the same, then it apply cleanly, if not, then you can solve the merge conflict.

I can drop ``the cherry-picked part from the commit``, if that's what you want.

PolynomialDivision:
Yeah I know. However, now github does not recognize it coming from packages repository. I think it is fine to merge it like this.

BKPepe:
So, you can merge it, right? ;-)

PolynomialDivision:
Yes.

PR #926, opened 2022-12-01
bird1-openwrt: change maintainer

mwarning:
Maintainer: @eloicaso
Description:
As discussed and agreed here https://github.com/openwrt/routing/pull/900, the maintainer will be changed from @eloicaso to @rogerpueyo

Please review and give a quick OK. Then we can merge.

### DIFF
--- a/bird1-openwrt/bird1-ipv4-openwrt/Makefile
+++ b/bird1-openwrt/bird1-ipv4-openwrt/Makefile
@@ -39,7 +39,7 @@ define Package/$(uci)
   SECTION:=net
   CATEGORY:=Network
   SUBMENU:=Routing and Redirection
-  MAINTAINER:=Eloi Carbo <eloicaso@openmailbox.org>
+  MAINTAINER:=Roger Pueyo Centelles <roger.pueyo@guifi.net>
   URL:=https://github.com/eloicaso/bird-openwrt/
   DEPENDS:=+$(BIRD_PKG) +libuci +uci
 endef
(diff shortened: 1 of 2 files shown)


rogerpueyo (approved):

rogerpueyo (approved):

rogerpueyo:
Fantastic, thanks!

mwarning:
OK, need to figure out why the pipeline fails.

rogerpueyo:
> OK, need to figure out why the pipeline fails.

Could it be because of the Makefile nested location inside folders `bird1-openwrt/bird-ipv{4,6}-openwrt/`? I believe these are the only ones, while the rest of Makefile are directly under `package-folder/Makefile`

Anyway, the test logs say:
```
make[1]: Entering directory '/home/build/openwrt'
make[1]: *** No rule to make target 'package/bird1-openwrt/bird1-ipv6-openwrt/download'.  Stop.
make[1]: Leaving directory '/home/build/openwrt'
make: *** [/home/build/openwrt/include/toplevel.mk:224: package/bird1-openwrt/bird1-ipv6-openwrt/download] Error 2
```

There's no download defined for the packages, but the source does not need to be downloaded, it's right there in `./src/`.

rogerpueyo:
> OK, need to figure out why the pipeline fails.

https://github.com/openwrt/routing/pull/928/checks

mwarning:
yes, but I do not know how to fix this..
@aparcar any idea?

rogerpueyo:
Sorry, I wasn't very clear. What I meant was to merge both Makefiles (for IPv4 and IPv6 packages) into a single one and put it directly into the `bird1-openwrt` folder. I did so #928 and both the tests pass and the local compilation works.

But yes, maybe this is not needed if @aparcar can tweak the tests.

rogerpueyo:
Hi, @mwarning,

I wonder if you could merge this commit, even if the checks are not passing (in the end, it's just a name change). This way I can update a few pending pull requests, including the one that fixes the Makefile to pass the checks.

Thanks!

mwarning:
@aparcar thank you!

PR #938, opened 2023-01-22
bmx7: bump PKG_RELEASE for libiwinfo ABI change

Ansuel:
Bump PKG_RELEASE for libiwinfo ABI change for bmx7 package.

Signed-off-by: Christian Marangi <ansuelsmth@gmail.com>

### DIFF
--- a/bmx7/Makefile
+++ b/bmx7/Makefile
@@ -2,7 +2,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=bmx7
 PKG_VERSION:=7.1.1
-PKG_RELEASE:=4
+PKG_RELEASE:=5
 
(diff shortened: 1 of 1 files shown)


mwarning:
@axn please have a look
@Ansuel was the release tar ball changed? I do not understand how the change of libiwinfo needs a change of PKG_RELEASE here.

dhewg:
Public structs gained new members, so sizeof(thething) is bigger now.
It's all backwards compatible, bmx7 doesn't require any code changes, but it needs to be rebuilt so it's compatible. And we want the bmx7 package to have the updated libiwinfo20230121dependency. It could be more elegant, but as-is we need the bump.

It's like https://github.com/openwrt/openwrt/commit/ee47a28cec01c7943238bae45f65a98e4fc9abbe

PR #946, opened 2023-02-09
ci: sync with latest openwrt/packages

PolynomialDivision:
Sync with latest changes on openwrt/packages.

### DIFF
--- a/.github/workflows/check-autorelease-deprecation.yml
+++ b/.github/workflows/check-autorelease-deprecation.yml
@@ -28,6 +28,8 @@ jobs:
 
       - name: Determine changed packages
         run: |
+          RET=0
+
           # only detect packages with changes
           PKG_ROOTS=$(find . -name Makefile | \
             grep -v ".*/src/Makefile" | \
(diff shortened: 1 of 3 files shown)


BKPepe:
I wonder why this is sitting here for 3 weeks without this getting merged. I haven't noticed this PR before, but right now, it fails for you because of:
```
tar: can't open '/ci/*.ipk': No such file or directory
tar: invalid magic
tar: short read
```

And if you take a look at it closer, you will see that it tries to compile three packages from the packages repository as fallback. You can see it in CI under "Determine changed packages", where it tries to compile:
```Building vim attendedsysupgrade-common bmon```, you can check what I did in https://github.com/openwrt/routing/pull/954 , but there are some packages which are failing right now.

BKPepe:
No need to ping me, when you backported those changes just partially, and your PR is still failing. From my point of view, this is incomplete as it is now.

Those 3 checks... 
- Two of them would be fixed by this: https://github.com/openwrt/packages/pull/21327 
- The another one needs to be reported it is related to cjdns. Did you do it?

We can not merge it unless you are willing to do something about that, though.

PolynomialDivision:
PR now sucesfully. :)

BKPepe:
Ehmm... it seems like you havent read my comment. Too bad. Guys were pinging you multiple times and you did not anything then merging this unfinished PR.

I am disappointed. :-(

PR #961, opened 2023-03-23
cjdns: pass for dangling pointer warning

VenetiaFurtado:
Maintainer: @wfleurant 
Compile tested: builds for x86_64
Run tested: Build only testing

Description: same build error seen on my machine as seen in https://github.com/openwrt/routing/issues/958
Fix proposed is similar to https://github.com/openwrt/routing/issues/738

### DIFF
--- a/cjdns/Makefile
+++ b/cjdns/Makefile
@@ -81,7 +81,7 @@ define Build/Compile
 	CC="$(TARGET_CC)" \
 	AR="$(TARGET_AR)" \
 	RANLIB="$(TARGET_RANLIB)" \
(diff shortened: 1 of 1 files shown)


BKPepe:
We can proceed further, but the formality check fails for you. Would you mind to fix it, please?

wfleurant:
no objections

wfleurant:
@cjdelisle please advise if you're not too busy

cjdelisle:
Oh, that's a bug in cjdns. What version are you using? I can push a fix against that version.

VenetiaFurtado:
It looks like version `v21.1` is being used.
 https://github.com/VenetiaFurtado/openwrt-routing/blob/master/cjdns/Makefile#L20

cjdelisle:
I fixed the cause of the issue here: https://github.com/openwrt/routing/pull/964
Thanks for bringing this to my attention.

PR #977, opened 2023-06-12
olsrd: update to 2023-06-12

PolynomialDivision:
Update to latest version.

Maintainer: me
Compile tested: ramips
Run tested: ramips


PolynomialDivision:
I'm getting crazy with the CI. I don't understand why the PKG_MIRROR_HASH should be different for some targets. I tested that now multiple times on my machine and I will try to merge this now and see if it will still build.

BKPepe:
Thats very odd behavior and weird decision to blindly merge this. Why would you do that if this is here only for 5 Hours? Be patient

PolynomialDivision:
I strongly believe it is caused by:
https://github.com/openwrt/openwrt/commit/75ab064d2b38e70746af1718ed7cdbafb906249e

There was already some other fix:
https://github.com/openwrt/openwrt/commit/5f1758ef14575df4e86896526b1c2035c231899e

PolynomialDivision:
image

PolynomialDivision:
Okay, the tar is not working correctly on my host system. That was the error. I will merge this and than again cherry-pick latest CI things.

PolynomialDivision:
> Hm. cry You should really wait until CI is finished if you are really unsure what are you doing or what was wrong.

The CI was already finished for the PR. I just deleted the rebase on the latest ci changes. Look into the IRC, there you can find the explanation what was wrong. It now builds fine except on my local machine, however, only if not already the tar.xz is uploaded to openwrt mirror.

BKPepe:
> Look into the IRC, there you can find the explanation what was wrong.

That's on you here to add a reference to the IRC history. If you notice correctly, I am not been using IRC for a few months already and have not even connected there. There are many channels to use - GitHub, IRC, mailing lists, OpenWrt forum, etc. Too many ways. For contributors, it is easier to use GitHub. Just my two cents.

Maybe in the future, I will start using IRC again.

PolynomialDivision:
@BKPepe My host system is somehow corrupt. When it packages the source to tar.xz something fails. @jow- said that the checksum field differs between the "openwrt docker version" and my local build machine. That's why I just used the PKG_MIRROR_HASH of the "docker build" and the CI was fine with it. However, I have no idea why I am the only one having those issues.

Important code:
- https://git.openwrt.org/?p=openwrt/openwrt.git;a=blob;f=tools/tar/patches/100-symlink-force-root-name.patch
- https://git.openwrt.org/?p=openwrt/openwrt.git;a=blob;f=tools/tar/patches/110-symlink-force-permissions.patch

PR #978, opened 2023-06-14
[23.05] olsrd: update to 2023-06-12

PolynomialDivision:
Update to latest version.

Remove upstreamed patch:
- 100-rename-avl-to-olsrd_avl.patch

(cherry picked from commit aab3b64e09a8c7969fdd9b4a793fda955d4fcaf5)

Maintainer: me
Compile tested: mt7622
Run tested: mt7622


BKPepe:
You should wait more time on this one. I've checked how it went with https://github.com/openwrt/routing/pull/977 and I see that the CI failed for several targets.

E.g.
```
WARNING: PKG_MIRROR_HASH does not match olsrd-2023-06-12-a9b3f1ac.tar.xz hash d8c80228b4a79fed12d6f967008da1e54ef893e6b252a394597907b2796ef211
```

Most likely it will happen on buildbots, too!

PolynomialDivision:
@BKPepe What you cite are old error messages. The CI here just fails because of the QEMU testing, that is somehow broken. You can see the same behavior in https://github.com/openwrt/packages. Now you already merged the gpsd update here: https://github.com/openwrt/packages/commits/openwrt-23.05 If we don't do this backport the 23.05 buildbots will fail to build olsrd.

BKPepe:
No, you get it wrong. You should wait in the first time to let the CI finish its stuff, otherwise, it is misleading as hell. There is no obvious or clear way to know that the CI passed for the latest commit. In the master branch, there are no checks. In the pull request for the master branch, you do have only a few checks passed and some of them were failing because of checksum.

You should be more patient, though. Calm down a little bit.

**What I was saying is that we are going to wait to see if those checksums for 23.05 branch will pop out too or not.** They did not.  Usually, cherry-picks to the stable branch should be done after 1-2 days.  You are still in a hurry and issues could happen.

Yes. I merged the stuff to 23.02 in the packages repository, because the CI was alright on the first sight (and as well on the master branch), which we could not say about the olsrd for the msater branch, right?

PR #992, opened 2023-07-18
babeld: update to 1.13

PolynomialDivision:
Announcement:
https://alioth-lists.debian.net/pipermail/babel-users/2023-July/004100.html

Babeld removed diversity routing [0], so we need to adjust ubus bindings. Further. we need to add a patch that makes local_kind accessible again: "100-local-make-local_kind-function-accessible.patch"

Refresh patches:
- 600-add-ubus.patch

[0] - https://github.com/jech/babeld/commit/a0816083356e5d33fb71e0e30d92aa3bf335d7ea

Maintainer: me
Compile tested: mt7621, mt7622
Run tested: mt7622

### DIFF
--- a/babeld/patches/100-local-make-local_kind-function-accessible.patch
+++ b/babeld/patches/100-local-make-local_kind-function-accessible.patch
@@ -1,0 +1,39 @@
+From b29cb705c3b717a7d5c61719936464438b9a48f0 Mon Sep 17 00:00:00 2001
+From: Nick Hainke <vincent@systemli.org>
+Date: Fri, 15 Jan 2021 15:01:31 +0100
+Subject: [PATCH] local: make local_kind function accessible
+
+The kind-definitions are contained in the header file:
+  LOCAL_FLUSH 0
+  LOCAL_ADD 1
+  LOCAL_CHANGE 2
+
+The function (local_kind) that converts them into strings again
+is private. New addons (for example openwrt ubus bindings) that make
+use of babeld should also use the same function to convert the kind
+to string format. This allows to ensure uniformity over all addons.
+
+Signed-off-by: Nick Hainke <vincent@systemli.org>
+---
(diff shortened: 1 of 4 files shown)


BKPepe:
You merged that within 40 minutes around midnight, in that time, trust me, no one will look at it. :-(

babeld/patches/100-local-make-local_kind-function-accessible.patch
```diff
@@ -0,0 +1,39 @@
+From b29cb705c3b717a7d5c61719936464438b9a48f0 Mon Sep 17 00:00:00 2001
+From: Nick Hainke <vincent@systemli.org>
+Date: Fri, 15 Jan 2021 15:01:31 +0100
+Subject: [PATCH] local: make local_kind function accessible
```
BKPepe, line 4: > Further. we need to add a patch that makes local_kind accessible again: "100-local-make-local_kind-function-accessible.patch"

Why do we need this patch? Did you try to send it to upstream? 

It looks like to me without further looking into it that you are changing something that is not exactly wanted by upstream.

PolynomialDivision: https://github.com/jech/babeld/commit/ef3eb65c8615b3dadab7f6f4e2e2383a71d0463f
https://github.com/jech/babeld/commit/fe9407b850dfda4e3c433edfa027bd1192732af8

BKPepe: Yes, I see that it is right now included in that repository, but you did not describe it anyhow in the OP. So, yes, it is better to include it instead of blindly adding patches, otherwise, the reviewer needs to spend more time if it is included in the upstream or it isnt. What I only found was this PR https://github.com/jech/babeld/pull/72 2 years ago.

PR #995, opened 2023-07-21
bird1: Remove, as EOL has been reached

marek22k:
Closes https://github.com/openwrt/routing/issues/993


BKPepe:
You should put to the commit description those links about deprecation, because no one by using git log will take a look what it was closing. :-)

marek22k:
I thought that was redundant because I already referenced the issue that links to it. However, I can still include the links if you like.

PR #996, opened 2023-07-21
quagga: Remove, as is no longer being developed

marek22k:
Closes https://github.com/openwrt/routing/issues/994


mwarning:
Can confirm, the sources are not available anymore.

PR #998, opened 2023-07-23
[23.05] nodogsplash: update to 5.0.1

mwarning:
Cherry picked from https://github.com/openwrt/routing/pull/997 since this includes an important security fix to protect against path tracersal attacks.

### DIFF
--- a/nodogsplash/Makefile
+++ b/nodogsplash/Makefile
@@ -6,12 +6,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=nodogsplash
-PKG_VERSION:=5.0.0
+PKG_VERSION:=5.0.1
 PKG_RELEASE:=1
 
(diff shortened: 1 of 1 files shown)


mwarning:
@bluewavenet @lynxis fyi

mwarning:
@PolynomialDivision hi, can you have a second look here? Thank you.

BKPepe:
Hmm, too bad that in the commit message, there are not explained changes like why they are additional dependencies, why the conflict was changed, etc. Because it looks like usual bump, which it isnt. :-(

mwarning:
Ah, I see. Yes, too bad.  For the record, the dependencies were missing.

mwarning:
@BKPepe there is probably no point in adding it to the commit now in this MR. If that's ok, then let's merge it.

PR #1002, opened 2023-07-27
nodogsplash: remove opennds from conflicts

mwarning:
OpenNDS lists nodogsplash a conflict as well.
This causes a circular reference that is not allowed.


mwarning:
@hnyman @bluewavenet  fyi

bluewavenet:
@mwarning 
We should increment PKG_RELEASE as well I guess.

PR #1005, opened 2023-07-29
opennds: Release v10.1.2

bluewavenet:

Compile tested: arm_cortex-a7_neon-vfpv4, mipsel_24kc, x86-64

Run tested: arm_cortex-a7_neon-vfpv4, mipsel_24kc, x86-64; on snapshot, 23.05, 22.03

Description:
opennds (10.1.2)

Security Advisory - This version contains fixes for multiple potential security vulnerabilities
Credit - Stanislav Dashevskyi - standash.github.io [standash]
It also contains some minor bug fixes
  * Fix - Generate unique sha256 faskey if not set in config - CVE-2023-38324 [bluewavenet]
  * Fix - NULL pointer dereference if user_agent is NULL - CVE-2023-38320, CVE-2023-38322 [bluewavenet]
  * Fix - NULL pointer dereference if authdir is called with an incomplete or missing query string - CVE-2023-38313, CVE-2023-38314, CVE-2023-38315 [bluewavenet]
  * Fix - remove deprecated and non-functioning unescape callback - CVE-2023-38316 [bluewavenet]
  * Fix - prevent potential recursive dependency and detect if conflicting package is installed  (in addition to the `CONFLICTS: nodogsplash` line in the makefile, code was added to check for the presence of nodogsplash on opennds startup) [bluewavenet]

### DIFF
--- a/opennds/Makefile
+++ b/opennds/Makefile
@@ -6,12 +6,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=opennds
-PKG_VERSION:=10.1.1
+PKG_VERSION:=10.1.2
 PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/opennds/opennds/tar.gz/v$(PKG_VERSION)?
-PKG_HASH:=60ce15f5aa96f7e7f3b239a0029f74c0ba900d3db72b209ba6e6d36a5bbef138
+PKG_HASH:=818c24a8704e584665b493857086bccc3c55629977841ddffc185f7d082ef279
 PKG_BUILD_DIR:=$(BUILD_DIR)/openNDS-$(PKG_VERSION)
 
 PKG_MAINTAINER:=Rob White <rob@blue-wave.net>
@@ -30,7 +30,6 @@ define Package/opennds
   DEPENDS:=+libmicrohttpd-no-ssl
   TITLE:=open Network Demarcation Service
   URL:=https://github.com/opennds/opennds
-  CONFLICTS:=nodogsplash
 endef
 
 define Package/opennds/description
@@ -66,7 +65,6 @@ define Package/opennds/install
 	$(CP) $(PKG_BUILD_DIR)/forward_authentication_service/PreAuth/theme_user-email-login-custom-placeholders.sh $(1)/usr/lib/opennds/
 	$(CP) $(PKG_BUILD_DIR)/forward_authentication_service/libs/get_client_interface.sh $(1)/usr/lib/opennds/
 	$(CP) $(PKG_BUILD_DIR)/forward_authentication_service/libs/client_params.sh $(1)/usr/lib/opennds/
-	$(CP) $(PKG_BUILD_DIR)/forward_authentication_service/libs/unescape.sh $(1)/usr/lib/opennds/
 	$(CP) $(PKG_BUILD_DIR)/forward_authentication_service/libs/authmon.sh $(1)/usr/lib/opennds/
 	$(CP) $(PKG_BUILD_DIR)/forward_authentication_service/libs/dnsconfig.sh $(1)/usr/lib/opennds/
 	$(CP) $(PKG_BUILD_DIR)/forward_authentication_service/libs/download_resources.sh $(1)/usr/lib/opennds/
--- a/opennds/Makefile
+++ b/opennds/Makefile
@@ -30,6 +30,7 @@ define Package/opennds
   DEPENDS:=+libmicrohttpd-no-ssl
   TITLE:=open Network Demarcation Service
   URL:=https://github.com/opennds/opennds
+  CONFLICTS:=nodogsplash
 endef
 
 define Package/opennds/description


bluewavenet:
@mwarning @PolynomialDivision @BKPepe 
This is a critical update fixing several potential security vulnerabilities as noted in the comment above.

PolynomialDivision:
Can you add a notice that you removed the conflict for nodogslpash in the commit message?

bluewavenet:
@PolynomialDivision 

> a notice that you removed the conflict

Implicitly, that is what this means:

> Fix - prevent potential recursive dependency and detect if conflicting package is installed [bluewavenet]

I have updated it ;-)

bluewavenet:
@PolynomialDivision @mwarning @BKPepe 
It is interesting that, as both opennds and nodogsplash have the common component `ndsctl` (albeit from a different code base), `opkg` detects that component is already present from a previous install of the "conflicting package" and fails the new install.
This works for both nodogsplash and opennds.

It is a different matter if the operating system is **_not_** OpenWrt, so some detection code has been added to opennds to mitigate that case.

BKPepe:
> Fix - prevent potential recursive dependency and detect if conflicting package is installed (explicitly, the CONFLICTS: nodogsplash line in the makefile was removed and code added to check for the presence of nodogsplash on opennds startup) [bluewavenet]

Still, I do think that there should be that conflict in the Makefile as we can not relay that it won't be changed in the future in your source code.  I haven't followed your recent discussion about nodogosplash closely.

opennds/Makefile
```diff
@@ -30,7 +30,6 @@ define Package/opennds
   DEPENDS:=+libmicrohttpd-no-ssl
   TITLE:=open Network Demarcation Service
   URL:=https://github.com/opennds/opennds
-  CONFLICTS:=nodogsplash
```
BKPepe, line 33: Okay, I found time to check out the recent discussion from here https://github.com/openwrt/routing/pull/997#issuecomment-1655227840

Unfortunately, the conflict is still required or if you want needed to have it at least in one package. Why? Both packages nodogsplash and openddns provide the same file. To be more specific:
- https://github.com/openwrt/routing/blob/967dde509b90c2166773ec672b2e8e8b81b6288a/nodogsplash/Makefile#L45
- https://github.com/openwrt/routing/blob/d038ab3d9e28a454221ce95e68b15dd3d70ed3c6/opennds/Makefile#L46

Even if you think that opkg is smart, actually, you can not have users have the option to have installed both of these packages. This can be easily reproduced and tested by using ``--force-overwrte``  to opkg. That it happens that you would have installed both packages and you would not know from which package you have that binary.

TL;DR: So, please, specify the conflict even if opkg does not allow to install it side by side, users will still be able to do that. It is always a good idea to avoid potentional issues. ;-)

BKPepe (changes requested):

opennds/Makefile
bluewavenet: @BKPepe 
If the opennds startup checks were removed and you force installed both, then nothing bad would happen, just portal logins would not work. You could still ssh to the router etc.
However the added startup checks in opennds will prevent it from starting even if force installed.

On a non-OpenWrt Linux, where in place compiling and installing is common, the startup checks are essential, but that is not relevant here.

So in my opinion the `CONFLICTS` entry is not needed.

But if you still think it is needed, no problem putting it back in.....

BKPepe: >If the opennds startup checks were removed and you force installed both, then nothing bad would happen, just portal logins would not work

I see, but still, I stand for my comment. You really don't want to have option to have installed two packages providing the same binary. If you start mixing stuff, it just calls for problems, which can be simply avoided. 

* I know, it is not your case, but it can be huge problem for SSL binaries, they do provide binary (same location, same name), but with different code base and also, the another good example is for full/small variants. That's why there are conflicts. It should be also used here. It was there, I don't see why it should be removed just now...

bluewavenet: The problem though is that the CONFLICTS will only be in one of them, whereas logically it should be in both, but we saw what happened when we did that. CONFLICTS does not work as expected.
So with it added back in, if opennds is installed you could install nodogsplash, if it wasn't for opkg stopping it because of ndsctl.
All this does show there is a larger issue here, with the build system.
Even with the CONFLICTS entry, you could still force-overwrite could you not?

Interesting but not worth discussing any further here.

I'll put it back in ;-)

BKPepe: When there is conflict, you can not force it to have it installed. Also, one conflict is enough.  ;)

bluewavenet:
Arrrgh

opennds/Makefile
bluewavenet: Ok, it's back in and ci is queued....

bluewavenet:
@BKPepe 
Thank you!

BKPepe:
No need to create PRs to stable branches. I will do cherry picks, there.

Dne po 31. 7. 2023 13:54 uživatel Rob White ***@***.***>
napsal:

bluewavenet:
@BKPepe 
You read my mind, I was about to do PRs for 23.05 and 22.03...

> I will do cherry picks

If you are sure you have time, that's excellent, thank you.

bluewavenet:
@BKPepe 
It has propagated more or less everywhere now in master.
I'm happy to do PR's for the cherry picks to stable branches to save you time....

PR #1023, opened 2023-10-09
nodogsplash: update to version 5.0.2

mwarning:
Fix a buffer overflow in case the query string is too long.

Maintainer: me
Compile tested: Nexx wt3020, ramips/mt7620
Run tested: save as compiled; tested internet blocked and unblocked after splash site click

### DIFF
--- a/nodogsplash/Makefile
+++ b/nodogsplash/Makefile
@@ -6,12 +6,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=nodogsplash
-PKG_VERSION:=5.0.1
-PKG_RELEASE:=2
+PKG_VERSION:=5.0.2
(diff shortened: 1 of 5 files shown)


nodogsplash/Makefile
```diff
@@ -40,6 +40,7 @@ define Package/nodogsplash/description
 endef
 
 define Package/nodogsplash/install
+	$(CP) ./files/* $(1)/
```
BKPepe, line 43: Uhh, this is not good.  The PR was draft and then it was merged. 
I would like to avoid this. You can not use CP for all. The previous solution was better, not the best.

mwarning: So you want each file to be referenced?

PR #1027, opened 2023-10-10
nodogsplash: explicit file copy

mwarning:
Maintainer: me
Compile tested: pipeline

### DIFF
--- a/nodogsplash/Makefile
+++ b/nodogsplash/Makefile
@@ -40,19 +40,23 @@ define Package/nodogsplash/description
 endef
 
 define Package/nodogsplash/install
-	$(CP) ./files/* $(1)/
+	$(INSTALL_DIR) $(1)/usr/lib/nodogsplash
+	$(INSTALL_BIN) ./files/usr/lib/nodogsplash/restart.sh $(1)/usr/lib/nodogsplash/
 	$(INSTALL_DIR) $(1)/usr/bin
 	$(INSTALL_BIN) $(PKG_BUILD_DIR)/nodogsplash $(1)/usr/bin/
 	$(INSTALL_BIN) $(PKG_BUILD_DIR)/ndsctl $(1)/usr/bin/
+	$(INSTALL_DIR) $(1)/etc/nodogsplash/htdocs/
+	$(INSTALL_DATA) $(PKG_BUILD_DIR)/resources/splash.html $(1)/etc/nodogsplash/htdocs/
+	$(INSTALL_DATA) $(PKG_BUILD_DIR)/resources/splash.css $(1)/etc/nodogsplash/htdocs/
+	$(INSTALL_DATA) $(PKG_BUILD_DIR)/resources/status.html $(1)/etc/nodogsplash/htdocs/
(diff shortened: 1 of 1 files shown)


mwarning:
@BKPepe how is this?

mwarning:
@BKPepe can I cherry-pick this into the other MRs for the other branches? Or do you want me to squash them?

mwarning:
@BKPepe ping

BKPepe:
I will look into it tomorrow.

čt 19. 10. 2023 v 18:52 odesílatel Moritz Warning ***@***.***>
napsal:

nodogsplash/Makefile
```diff
@@ -49,6 +48,10 @@ define Package/nodogsplash/install
 	$(INSTALL_DIR) $(1)/etc/init.d
 	$(INSTALL_DIR) $(1)/etc/uci-defaults
 	$(INSTALL_DIR) $(1)/usr/lib/nodogsplash
+	$(CP) ./files/etc/config/nodogsplash $(1)/etc/config/
+	$(CP) ./files/etc/init.d/nodogsplash $(1)/etc/init.d/
+	$(CP) ./files/etc/uci-defaults/40_nodogsplash $(1)/etc/uci-defaults/
+	$(CP) ./files/usr/lib/nodogsplash/restart.sh $(1)/usr/lib/nodogsplash/
```
BKPepe, line 54: This still does not look good. We are using simple CP. This needs to be reworked.

We should make it more clear, which will be more better to read this install section and it will be simple as it is.
1. Create folder, then install files to it and then create another folder and install there files, which are there.
The order if you do first binaries, then config files, does not matter. Important is that it is not lost. 

This is done for binary (a few lines above) and then you are creating three folders and it might be messy.

Anyway, I would suggest to use installation macros, which are listed here:
https://openwrt.org/docs/guide-developer/packages#file_installation_macros

E.g., INSTALL_CONF for config file, INSTALL_BIN for init config, etc.

This makes sure that it has permissions on the router, which we want and that it has not been tinkerered with.

nodogsplash/Makefile
```diff
@@ -40,19 +40,27 @@ define Package/nodogsplash/description
 endef
 
 define Package/nodogsplash/install
-	$(CP) ./files/* $(1)/
+	$(INSTALL_DIR) $(1)/usr
+	$(INSTALL_DIR) $(1)/usr/lib
```
BKPepe, line 44: I think, we can drop this, right? I bet that INSTALL_DIR is recursive, so leaving only /usr/lib/donogsplash should be enough. Same applies for /etc.

Otherwise, good job! :) I feel like we are reaching the end.

mwarning: Thank you for you patience. I have also removed most of the needless INSTALL_DIR calls.

BKPepe (approved):

nodogsplash/Makefile
BKPepe: Looks good now! Thank you and sorry that it took so long.

PR #1032, opened 2023-11-07
mcproxy: fix igmpv3 and mld record type

rthakur33:
This patch fixes that if IGMPv3 Membership Report packet on the LAN interface with ALLOW_NEW_SOURCES record containing specific source is sent, then IGMPv3 Membership Report packet is received on the WAN interface and group record is ALLOW_NEW_SOURCES for the group with a matching source list.

### DIFF
--- a/mcproxy/patches/0008-fix-igmpv3-mld-record-type.patch
+++ b/mcproxy/patches/0008-fix-igmpv3-mld-record-type.patch
@@ -1,0 +1,42 @@
+--- a/mcproxy/src/proxy/igmp_sender.cpp
++++ b/mcproxy/src/proxy/igmp_sender.cpp
+@@ -54,10 +54,14 @@ bool igmp_sender::send_record(unsigned int if_index, mc_filter filter_mode, cons
+         m_sock.leave_group(gaddr, if_index);
+         return true;
+     } else if (filter_mode == EXCLUDE_MODE || filter_mode == INCLUDE_MODE) {
+-        m_sock.join_group(gaddr, if_index);
+         std::list<addr_storage> src_list;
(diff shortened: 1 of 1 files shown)


mwarning:
@sbyx ping

BKPepe:
Is this patch included in upstream? If not, are you willing to send it there?

rthakur33:
> Is this patch included in upstream? If not, are you willing to send it there?

@BKPepe , thanks for your comment, I actually tried there first https://github.com/mcproxy/mcproxy/pull/18 but got no response hence raised this as a patch here.

BKPepe:
I was thinking, if the upstream is dead for more than 7 years. What about rather dropping this package at all? Here are a few points which leads that removing is gonna be helpful:

- If there is any issue (bug/security vulnerability), upstream developers probably will not look into it.
- If you want to have fixed something there, then most likely, it will not get any attention as well.
- I am not able to see this package in other GNU/Linux distributions such as Debian, Ubuntu, etc.

Sometimes, it is just better to move on instead of relying on something which is obviously not working.

rthakur33:
> I was thinking, if the upstream is dead for more than 7 years. What about rather dropping this package at all? Here are a few points which leads that removing is gonna be helpful:
> 
> * If there is any issue (bug/security vulnerability), upstream developers probably will not look into it.
> * If you want to have fixed something there, then most likely, it will not get any attention as well.
> * I am not able to see this package in other GNU/Linux distributions such as Debian, Ubuntu, etc.
> 
> Sometimes, it is just better to move on instead of relying on something which is obviously not working.

+1 from me.

BKPepe:
I will create PR shortly.

Issue #1036, opened 2023-12-21
batman-adv: build error: redefinition of 'skb_vlan_eth_hdr'


AcidSlide:
Maintainer: @ecsv 
Environment: ramips / mipsel_24kc (mt76x8),  Xiaomi 4A 100M, openwrt-snapshot-4a2ff73

Description:

Error building/compiling **batman-adv**
```
In file included from <command-line>:
~/openwrt/build_dir/target-mipsel_24kc_musl/linux-ramips_mt76x8/batman-adv-2023.3/compat-hacks.h:38:35: error: redefinition of 'skb_vlan_eth_hdr'
   38 | static inline struct vlan_ethhdr *skb_vlan_eth_hdr(const struct sk_buff *skb)
      |                                   ^~~~~~~~~~~~~~~~
In file included from ~/openwrt/build_dir/target-mipsel_24kc_musl/linux-ramips_mt76x8/batman-adv-2023.3/compat-hacks.h:33:
./include/linux/if_vlan.h:66:35: note: previous definition of 'skb_vlan_eth_hdr' with type 'struct vlan_ethhdr *(const struct sk_buff *)'
   66 | static inline struct vlan_ethhdr *skb_vlan_eth_hdr(const struct sk_buff *skb)
      |                                   ^~~~~~~~~~~~~~~~
```

raenye:
Same here, with a different arch, target, and kernel (`mediatek/fiilogic` `linux-6.1.69`).

```
$ make package/feeds/routing/batman-adv/compile V=s
...
make[3]: Entering directory '/home/user/git/openwrt/build_dir/target-aarch64_cortex-a53_musl/linux-mediatek_filogic/linux-6.1.69'
  CC [M]  /home/user/git/openwrt/build_dir/target-aarch64_cortex-a53_musl/linux-mediatek_filogic/batman-adv-2023.3/net/batman-adv/bat_algo.o
In file included from <command-line>:
/home/user/git/openwrt/build_dir/target-aarch64_cortex-a53_musl/linux-mediatek_filogic/batman-adv-2023.3/compat-hacks.h:38:35: error: redefinition of 'skb_vlan_eth_hdr'
   38 | static inline struct vlan_ethhdr *skb_vlan_eth_hdr(const struct sk_buff *skb)
      |                                   ^~~~~~~~~~~~~~~~
[...]

PR #1039, opened 2023-12-23
openwrt-23.05: batman-adv: compat: Fix skb_vlan_eth_hdr conflict in stable kernels

ecsv:
Maintainer:  @simonwunderlich 
Compile tested: ramips / mipsel_24kc (mt76x8)
Run tested: -

Description:

The newest Linux stable kernel releases:

* v5.10.205
* v5.15.144
* v6.1.69

received a backported version of commit 1f5020acb33f ("net: vlan: introduce skb_vlan_eth_hdr()"). batman-adv must therefore not provide this function any longer for these kernel versions.

Closes: #1036

### DIFF
--- a/batman-adv/Makefile
+++ b/batman-adv/Makefile
@@ -4,7 +4,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=batman-adv
 PKG_VERSION:=2023.1
-PKG_RELEASE:=4
+PKG_RELEASE:=5
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://downloads.open-mesh.org/batman/releases/batman-adv-$(PKG_VERSION)
--- a/batman-adv/src/compat-hacks.h
+++ b/batman-adv/src/compat-hacks.h
@@ -28,7 +28,10 @@ static inline u32 batadv_get_random_u32_below(u32 ep_ro)
 
 #endif /* LINUX_VERSION_IS_LESS(6, 2, 0) */
 
-#if LINUX_VERSION_IS_LESS(6, 4, 0)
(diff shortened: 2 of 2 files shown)


herbetom:
From what i can tell is openwrt-23.05 still on 5.15.139:

https://github.com/openwrt/openwrt/blob/59fd8f08cf544a7f0fcaf700ca9af9292de3adcc/include/kernel-5.15

And the commit as mentioned in the PR description only backported since v5.15.144.

https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/include/linux/if_vlan.h?h=v5.15.144&id=3c4dcfbff15bae6a4da8ef5682d94e849612b47c

But I don't want to rule out the possibility that I'm missing something.

ecsv:
> From what i can tell is openwrt-23.05 still on 5.15.139:
> 
> https://github.com/openwrt/openwrt/blob/59fd8f08cf544a7f0fcaf700ca9af9292de3adcc/include/kernel-5.15
> 
> And the commit as mentioned in the PR description only backported since v5.15.144.
> 
> https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/include/linux/if_vlan.h?h=v5.15.144&id=3c4dcfbff15bae6a4da8ef5682d94e849612b47c
> 
> But I don't want to rule out the possibility that I'm missing something.

What are you trying to tell us with this information? Do you assume that openwrt-23.05.x will always stay on a kernel <  v5.15.144 and that we should there now remove this modification again? And even if OpenWrt is really neglecting 23.05.x, don't you think that at least some downstream projects should (and will) update to a newer v5.15.x bugfix release?

Or did you find a bug because of this change and we should therefore remove it again?

Or?

herbetom:
sorry, while trying to figure out if this change is relevant i mixed up the change here and the upstream change as i didn't look closely enough what the actual change beeing made here was and assumed something. So everyhing should be fine.

PR #1040, opened 2024-02-02
batman-adv: update packages to version 2024.0

ecsv:
Maintainer: @simonwunderlich 
Compile tested: x86_64
Run tested: x86_64

Description:

batman-adv
==========

* support latest kernels (4.19 - 6.8)
* add stateless multicast packet format support

batctl
======

* add stateless multicast packet format support
* bugs squashed:
  - Fix various length checks in tcpdump-like subcommand


batman-adv/patches/0003-Revert-batman-adv-Switch-to-linux-sprintf.h.patch
```diff
@@ -0,0 +1,93 @@
+From: Sven Eckelmann <sven@narfation.org>
+Date: Fri, 2 Feb 2024 22:49:00 +0100
+Subject: Revert "batman-adv: Switch to linux/sprintf.h"
+
+This reverts commit f0fb49c5ab70dfa064f0aa8d1c5d84f65e8cbc86.
```
BKPepe, line 5: It would be great, if we do have reason, why do we need to have this patch.
Okay, it reverts something without any specific details.

BKPepe:
Thanks! :clap:

PR #1043, opened 2024-02-03
openwrt-22.03: treewide: remove AUTORELEASE

ecsv:
Automatically compute and substitute current values for all `$(AUTORELEASE)` instances as this feature is deprecated and shouldn't be used. Since commit 2584b0cecae ("CI: deprecate $(AUTORELEASE) via comments"), this is even announced by the CI

The following temporary change was made to the core:

    diff --git a/rules.mk b/rules.mk
    index 57d7995d4fa8..f16367de87a8 100644
    --- a/rules.mk
    +++ b/rules.mk
    @@ -429,7 +429,7 @@ endef
     abi_version_str = $(subst -,,$(subst _,,$(subst .,,$(1))))

     COMMITCOUNT = $(if $(DUMP),0,$(call commitcount))
    -AUTORELEASE = $(if $(DUMP),0,$(call commitcount,1))
    +AUTORELEASE = $(if $(DUMP),0,$(shell sed -i "s/\$$(AUTORELEASE)/$(call commitcount,1)/" $(CURDIR)/Makefile))

     all:
     FORCE: ;

And this command used to fix affected packages:

    ./scripts/feeds install -a routing
    for i in $(cd feeds/routing; git grep -l 'PKG_RELEASE:=.*AUTORELEASE' | \
                                  sed 's^.*/\([^/]*\)/Makefile^\1^';);
    do
      make package/${i%/Makefile}/download
    done


BKPepe:
Even though, the version is the same (I checked pimbd only...), I would keep the autorelease for OpenWrt 22.03. My point of view for this is that we dont want to force to install new version of packages if we miss some number that it is not identical just because of some deprecated function, which works. Thats my opinion on this.

PR #1060, opened 2024-04-07
cjdns: switch version to be APK compatible

aparcar:
Maintainer: @wfleurant 
Compile tested: mipsel_24kc
Run tested: -

Description
With a switch to APK, the version schema becomes a hard requirement and may not contain arbitrary letters. Remove the `v` and attach it only while downloading/building.

### DIFF
--- a/cjdns/Makefile
+++ b/cjdns/Makefile
@@ -17,13 +17,13 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=cjdns
-PKG_VERSION:=v21.1
+PKG_VERSION:=21.1
 PKG_RELEASE:=5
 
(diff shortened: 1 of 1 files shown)


cjdns/Makefile
```diff
@@ -17,13 +17,13 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=cjdns
-PKG_VERSION:=v21.1
+PKG_VERSION:=21.1
 PKG_RELEASE:=5
```
BKPepe, line 21: We might want to reset it to 1

aparcar: Isn't it more like r6 then?

BKPepe: Could be, but still I think that adding r should be done via treewide commit

wfleurant:
@aparcar thanks

PR #1062, opened 2024-04-11
treewide: upgrade ucitrack handling to use json

feckert:
Maintainer: @stargieg @PolynomialDivision @wfleurant 
Compile tested: no
Run tested: no

Description:
This depends on  [pullrequest](https://github.com/openwrt/luci/pull/7056) in the `LuCI` feed.

The `ucitrack` configuration under `/etc/config` defines which service must be restarted on a `LuCI` change. This `uci`  file defines how all this works together. Additionally there are `uci-defaults` scripts in the various applications that add additional configurations parameters to this `ucitrack` file if a service must be reloaded/restarted if a `uci` change was made in this application.

There are some problems with the current implementation:
* The `uci` should be used to configure the system and not for this kind of reload/restart dependency handling on `LuCI` changes.
*  On a system update with configuration restore of the `uci` file `ucitrack` the new behavior on `LuCI` reload/restart could not take into account because the new file is not be used.


feckert:
The [base PR](https://github.com/openwrt/luci/pull/7056) has been merged.
Since I have no commit rights for this feed.
Can one of those @BKPepe @aparcar who have this merge this change?

PR #1071, opened 2024-06-23
batman-adv: update packages to version 2024.2

ecsv:
Maintainer: @simonwunderlich 
Compile tested: x86_64
Run tested: x86_64

Description:

batman-adv
==========

* support latest kernels (4.19 - 6.10)

### DIFF
--- a/alfred/Makefile
+++ b/alfred/Makefile
@@ -3,12 +3,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=alfred
-PKG_VERSION:=2024.1
+PKG_VERSION:=2024.2
 PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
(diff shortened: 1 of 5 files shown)


simonwunderlich:
Thank you! Just for the record, the build server had some issues with downloading on the first two builds as open-mesh.org had some temporary problems, but they are fixed.

BKPepe:
I wonder, why you wouldnt retrigger the CI/CD before merging this?

ecsv:
I can't. I can create a status badge but not trigger a rebuild. I would have to do again a useless commit and force-push again the "no-change".

BKPepe:
Even though it might not seem like a good way to do it, it is still better to retrigger it like this so we are sure that those temporary issues are gone and that the CI/CD passes. Also, there is another option to ask someone with the commit permission to retrigger it. I did it after the merging, but it seems that CI/CD still can not download it. 

Right now, it looks like we do have CI/CD, and it complains about one issue, but we are ignoring it as we can. It is not always a great idea, though.

PR #1073, opened 2024-07-09
For mcproxy fixes

mhusaam:
Various improvements and fixes to the mcproxy package


BKPepe:
I wonder, those patches, which you would like to add, where are they from? Because they are not part of the upstream repository https://github.com/mcproxy/mcproxy not even they are in pending pull requests.

BKPepe:
Closing, we can not accept and merge so many patches as it leads to maintenance burden. It should be discussed with upstream.

PR #1074, opened 2024-07-20
mesh11sd: update to version 4.1.0

bluewavenet:

Compile tested: All

Run tested: arm_cortex-a7_neon-vfpv4, mipsel_24kc, mips_24kc, aarch64_cortex-a53;
    On 23.5 and master/snapshot.

Description: mesh11sd (4.1.0)
This release adds new functionality that includes support for:
 * Opportunistic Wireless Encryption (OWE) on mesh gates.
 * Customer[Client] Premises Equipment mode (CPE) on mesh gates.

Details can be found here:
https://github.com/openNDS/mesh11sd/releases/tag/v4.1.0

### DIFF
--- a/mesh11sd/Makefile
+++ b/mesh11sd/Makefile
@@ -8,7 +8,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=mesh11sd
-PKG_VERSION:=4.0.1
+PKG_VERSION:=4.1.0
 PKG_RELEASE:=1
 
 PKG_MAINTAINER:=Rob White <rob@blue-wave.net>
@@ -17,7 +17,7 @@ PKG_LICENSE_FILES:=LICENSE
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/opennds/mesh11sd/tar.gz/v$(PKG_VERSION)?
-PKG_HASH:=8029a0e41487d322b1bb20df05fc3ef2073b1239977f2b430b20fe7cabfe71de
+PKG_HASH:=80406e70fadf58320ec4891c8fb1e93d118d2927de7c6d9749a15da6768b4ea5
 PKG_BUILD_DIR:=$(BUILD_DIR)/mesh11sd-$(PKG_VERSION)
 
 include $(INCLUDE_DIR)/package.mk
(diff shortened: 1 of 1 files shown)


mesh11sd/Makefile
```diff
@@ -17,7 +17,7 @@ PKG_LICENSE_FILES:=LICENSE
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://codeload.github.com/opennds/mesh11sd/tar.gz/v$(PKG_VERSION)?
-PKG_HASH:=8029a0e41487d322b1bb20df05fc3ef2073b1239977f2b430b20fe7cabfe71de
+PKG_HASH:=80406e70fadf58320ec4891c8fb1e93d118d2927de7c6d9749a15da6768b4ea5
 PKG_BUILD_DIR:=$(BUILD_DIR)/mesh11sd-$(PKG_VERSION)
```
BKPepe, line 21: I know that it is not related to your PR, apologies for nitpicking, but isnt this already default?

bluewavenet: @BKPepe 
Default? Sorry, I don't follow.
But it is a new hash.... Does not the build system calculate the hash from the downloaded source tar.gz and compare what it gets with this value?
I test building the package using the SDK and it requires this to be updated.

BKPepe: This is what I mean:

```suggestion
```

bluewavenet: I see! Are you sure it is a default?

Shall we try it and see? It's Saturday, lets live dangerously :-D

BKPepe: Yup, its default https://github.com/openwrt/openwrt/blob/66177c081f1b05e4eb7f6c23d97954e8172c4336/include/package.mk#L11 , no need to specify the same here. :-)

bluewavenet: > The best line of code is no line of code

bluewavenet:
@BKPepe 
Interesting, Test Formalities fails....

BKPepe:
Thats not interesting. 😅 Its expected. All commits subjects in OpenWrt repositories must begin with prefix. In this case prefix is package name. It means: ``mesh11sd: foo``.

bluewavenet:
@BKPepe 

> Its expected.

I know that ;-)

bluewavenet:
@BKPepe 
I think we are good to go?

bluewavenet:
@BKPepe 
Thank you!
PS. Nitpicking is good ;-)

PR #1085, opened 2024-11-13
olsrd: fix APK package build

pktpls:
Maintainer: @PolynomialDivision hi :)
Compile tested: x86_64
Run tested: x86/64

Description:

Can't use the iptables meta packages apparantly:

```
ERROR: unable to select packages:
  ip6tables (virtual):
    note: please select one of the 'provided by'
          packages explicitly
    provided by: ip6tables-nft
                 ip6tables-zz-legacy
    required by: world[ip6tables]
                 olsrd-2024.06.09~d72be9ad-r1[ip6tables]
  iptables (virtual):
    note: please select one of the 'provided by'
          packages explicitly
    provided by: iptables-nft iptables-zz-legacy
    required by: olsrd-2024.06.09~d72be9ad-r1[iptables]
```

### DIFF
--- a/olsrd/Makefile
+++ b/olsrd/Makefile
@@ -34,7 +34,7 @@ endef
 define Package/olsrd
   $(call Package/olsrd/template)
   MENU:=1
-  DEPENDS:=+libpthread +libubus +iptables +IPV6:ip6tables
+  DEPENDS:=+libpthread +libubus
 endef
 
 define Package/olsrd/conffiles


pktpls:
Please backport to openwrt-24.10 branch as well, just in case it's decided to activate APK there as well :heart:

PolynomialDivision:
But we wanted the user to be able to select the correct dependency themselves? If someone wants to use iptables, they should do. Maybe deleting the dependency completely would be a solution?

pktpls:
> But we wanted the user to be able to select the correct dependency themselves? If someone wants to use iptables, they should do. Maybe deleting the dependency completely would be a solution?

Yes you're probably right, better to not force iptables-nft. I'll update it

PolynomialDivision:
I changed the author to myself to ensure a valid sign-off, with approval from @pktpls. It seems the real name discussion is still ongoing: https://github.com/openwrt/actions-shared-workflows/commit/12d9551f2d07ec34ac813da8612c8014fb393af6

pktpls:
This PR is not needed, sorry, I was using it wrong :grimacing: 

If any packages declares iptables/ip6tables as its dependency, then *something* needs to select a package providing it. I just selected iptables-nft and ip6tables-nft further up the dependency tree, and it's now fine.

The change with APK is not that we can't depend on virtual packages anymore, it's apparantly that there's no default provider selection anymore.

PR #1088, opened 2024-11-15
Add configuration to support MD5 auth with bird2

### LINKED ISSUE #1084: bird2: does not work with MD5 password
Please make sure that the issue subject starts with `<package-name>: `

Also make sure that the package is maintained in this repository and not in OpenWrt-base, OpenWrt-packages or OpenWrt-LuCI.

Issues related to releases below 18.06 and forks are not supported or maintained and will be closed.

# Issue template (remove lines from top till here)

Maintainer: @tohojo 
Environment: Banana pi - R4, OpenWRT snapshot

Description:

When using bird2 with an MD5 password to connect to peer we get the following error:
```
Socket error: Kernel does not support TCP MD5 signatures
```

It is because the kernel is not build with the option `CONFIG_TCP_MD5SIG`.

I saw that it has been solved on FRR in the following PR: https://github.com/openwrt/packages/pull/24178
I don't know if it is possible to do it the same in this repository. If yes, I can purpose the PR.

guillomep:
Maintainer: @tohojo 
Compile tested: x86-64, generic, snapshot
Run tested: x86-64, generic, snapshot

Description:
Fix #1084

Same correction as in FRR https://github.com/openwrt/packages/pull/24178

### DIFF
--- a/bird2/Makefile
+++ b/bird2/Makefile
@@ -8,7 +8,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=bird2
 PKG_VERSION:=2.15.1
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE:=bird-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=ftp://bird.network.cz/pub/bird
@@ -36,6 +36,7 @@ define Package/bird2
   SECTION:=net
   CATEGORY:=Network
   SUBMENU:=Routing and Redirection
+  KCONFIG:=CONFIG_TCP_MD5SIG=y
   DEPENDS:=+libpthread
   CONFLICTS:=bird1-ipv4 bird1-ipv6 bird4 bird6
 endef


guillomep:
> So this adds a dependency on the kmod package? Isn't that a little overkill to have as a hard dependency? It's perfectly possible to use bird without that. 

I haven’t found the kmod that allows to load this functionality but if there is one then this PR is useless.

> How many extra packages does it pull in when installing bird on a fresh install after this change?

I truly don’t know how to find that number

tohojo:
On 16 November 2024 00:13:25 CET, Guillaume Perrin ***@***.***> wrote:
>> So this adds a dependency on the kmod package? Isn't that a little overkill to have as a hard dependency? It's perfectly possible to use bird without that. 
>
>I haven’t found the kmod that allows to load this functionality but if there is one then this PR is useless.

If it's not a kmod, what does this patch do, exactly? AFAOU, bird and the kernel aren't built together, so how does setting a kconfig option in a package Makefile work, exactly?

>> How many extra packages does it pull in when installing bird on a fresh install after this change?
>
>I truly don’t know how to find that number

I would just install the package and see what dependencies it pulls in with and without this patch...

PolynomialDivision:
Do you want something like this?

https://github.com/openwrt/openwrt/commit/c8d5abd62b70137c70bf12e83b6d0708b980abb3

guillomep:
Yes this is probably a better way to do it that what I purpose and what has been done in FRR

PR #1090, opened 2024-12-06
bird2: Prevent unaligned access on ARM

tohojo:
Import patch from Freifunk to enable -mno-unaligned-access - see upstream bug report:

http://trubka.network.cz/pipermail/bird-users/2024-December/017944.html

Ref: https://github.com/freifunk-berlin/falter-packages/pull/454

### DIFF
--- a/bird2/Makefile
+++ b/bird2/Makefile
@@ -8,7 +8,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=bird2
 PKG_VERSION:=2.15.1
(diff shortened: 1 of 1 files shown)


BKPepe:
I think, this requires to bump PKG_RELEASE.

Dne pá 6. 12. 2024 11:47 uživatel Toke Høiland-Jørgensen <
***@***.***> napsal:

tohojo:
Ah yes, of course!

PR #1094, opened 2024-12-12
bird2: use official example config file

sch-m:
Maintainer: @bradfordzhang @tohojo 
Compile tested: x86_64, latest master
Run tested: x86_64, latest master

Description:
Using the openwrt example config (which seems to be a clone of the official example2 config) leads to the following error:

    bird: /etc/bird.conf:138:15 syntax error, unexpected '/'

In addition, the previous configuration seems to me to be too case-specific and cannot really be used as a template for a fresh configuration.


tohojo (approved):
Sure, sounds reasonable

PR #1097, opened 2024-12-17
bird2: update to 2.16.1

PolynomialDivision:
Previously, we used the -mno-unaligned-access flag to instruct GCC to forbid unaligned memory access on arm processors. This was a workaround for alignment issues that caused crashes.

This commit imports a patch from the BIRD mailing list discussion [0] that resolves the alignment issue by modifying the net_addr structure, removing the need for the GCC flag. The patch addresses the alignment problem more efficiently and avoids the performance, code size, and hardware optimization drawbacks of the flag.

[0] - http://trubka.network.cz/pipermail/bird-users/2024-December/017957.html

Maintainer: @tohojo 
Compile tested: arm7
Run tested: arm7

### DIFF
--- a/bird2/Makefile
+++ b/bird2/Makefile
@@ -7,12 +7,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=bird2
-PKG_VERSION:=2.16
+PKG_VERSION:=2.16.1
 PKG_RELEASE:=1
 
 PKG_SOURCE:=bird-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://bird.network.cz/download/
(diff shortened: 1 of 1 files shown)


tohojo (changes requested):
Needs PKG_RELEASE bump...

spolack:
Seems it was fixed upstream...
https://gitlab.nic.cz/labs/bird/-/commit/161aef353a56532752d5d0f97e432c0915058050

tohojo:
There are two patches upstream, let's include both - cf: http://trubka.network.cz/pipermail/bird-users/2024-December/017988.html

PolynomialDivision:
@tohojo I updated now to latest bird2 release.

tohojo:
Great, thanks!

PR #1098, opened 2024-12-18
bird3: add new bird3 package

PolynomialDivision:
@tohojo Is it fine if I add this package and list both you and me as maintainers?


tohojo:
Sure! Thanks for taking care of it :)

PR #1109, opened 2025-04-09
cjdns: patch of gyp imports support python3.11

ntbowen:
This solution:

First attempts to import MutableSet from collections.abc (Python 3.10+ location) Falls back to using collections.MutableSet for older Python versions Changes the class inheritance to use the properly imported MutableSet rather than directly referencing collections.MutableSet This approach ensures compatibility across different Python versions while maintaining the intended functionality of the OrderedSet class in the gyp build system.

Description:


wfleurant:
we need to 3-way diff a new patch because this also exists

https://github.com/openwrt/routing/blob/master/cjdns/patches/040-gyp-python_310.patch

please rename your source branch from 'master' to something like 'fix-import-etc...' master into master may not be accepted in review

did this compile / test well?

PR #1128, opened 2025-06-16
bird{2,3}: improve Makefile

BKPepe:
Maintainer: @tohojo , @PolynomialDivision

This is just a few things, which I modify, before these two packages can be moved to the packages feed.

### DIFF
--- a/bird2/Makefile
+++ b/bird2/Makefile
@@ -22,17 +22,9 @@ PKG_BUILD_DIR:=$(BUILD_DIR)/bird-$(PKG_VERSION)
 
 include $(INCLUDE_DIR)/package.mk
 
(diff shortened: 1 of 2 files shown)


BKPepe:
It looks like more packages are using INSTALL_DATA instead of INSTALL_CONF for configuration files. I will do treewide commit for it. :)

PolynomialDivision:
Nice thanks!

tohojo (approved):

BKPepe:
> It looks like more packages are using INSTALL_DATA instead of INSTALL_CONF for configuration files. I will do treewide commit for it. :)

👉 https://github.com/openwrt/routing/pull/1129

PR #1131, opened 2025-06-17
hnetd: remove it as the development is stalled

BKPepe:
Development of hnetd is stalled [1].
Last commit in the upstream repository is 7 years ago and currently, hnetd is not compiled in snapshot builds due to errors from GCC 14 [2]

[1] https://github.com/sbyx/hnetd/issues/67
Fixes: https://github.com/openwrt/routing/issues/1119 [2]

cc: @aparcar, @sbyx


BKPepe:
I think the link [1] says it all. 😇 😇  What else do you need? Issue was created in 2021, no new commits. Always, you can drop it and once, the issue is solved, you can re-add it. It's master branch. You will see if someone is using or not, if not, then nice. If yes, someone should step it and take the maintainership of hnetd.

I would not complicate it.

PR #1140, opened 2025-09-19
treewide: fix maintainer naming by removing incorrect quotation marks

feckert:
Maintainer: @BKPepe
Compile tested: no
Run tested: no

Description:

The extra quotation marks in the PKG_MAINTAINER line of the '.packageinfo'   file does not produce valid JSON when [package-metadata.pl pkgmanifestjson](https://github.com/openwrt/openwrt/blob/main/scripts/package-metadata.pl#L820) is used.

Therefore, the word with the surrounding quotation marks is removed.


feckert:
@BKPepe done

PR #1173, opened 2026-05-08
bird3: add protocol-specific build variants and Config.in

PolynomialDivision:
Bird's binary size is dominated by compiled-in routing protocols. This adds size-optimized package variants built with --with-protocols so users on flash-constrained devices can install only what they need without recompiling.

New packages:
- bird3-bgp:   BGP, RPKI, BFD + infrastructure (mrt, pipe, static)
- bird3-ospf:  OSPF, BFD + infrastructure
- bird3-babel: Babel + infrastructure
- bird3-mesh:  Babel, BGP, RPKI, BFD + infrastructure, for mesh
               networks using BGP-based overlay services (e.g. hostname
               distribution via bgpdisco)

All variants PROVIDE:=bird3 so existing DEPENDS:=+bird3 continue to work. They conflict with each other since all install /usr/sbin/bird.

For users building from source, Config.in adds per-protocol checkboxes under the bird3 menu in menuconfig. Protocols default to enabled to preserve existing behaviour, builders on tight flash budgets can uncheck what they do not need. PKG_CONFIG_DEPENDS ensures bird3 is rebuilt when the selection changes.

Maintainer: me, @tohojo 
Compile tested: tbd
Run tested: tbd

### DIFF
--- a/bird3/Config.in
+++ b/bird3/Config.in
@@ -1,0 +1,80 @@
+if PACKAGE_bird3
+
+comment "Routing protocols"
+comment "Protocols mrt, pipe, and static are always included"
+
+config BIRD3_PROTO_BGP
+	bool "BGP"
+	default y
+	help
+	  Border Gateway Protocol. Required for internet routing,
+	  route servers, and route reflectors.
+
+config BIRD3_PROTO_RPKI
+	bool "RPKI"
+	default y
+	help
+	  Route Origin Validation via the Resource Public Key
+	  Infrastructure. Used alongside BGP to validate route origins.
+
+config BIRD3_PROTO_BFD
+	bool "BFD"
+	default y
+	help
+	  Bidirectional Forwarding Detection. Provides fast link failure
+	  detection for BGP and OSPF sessions.
+
+config BIRD3_PROTO_OSPF
+	bool "OSPF"
+	default y
+	help
+	  Open Shortest Path First (OSPFv2 for IPv4, OSPFv3 for IPv6).
+	  Common interior gateway protocol for enterprise and ISP networks.
+
+config BIRD3_PROTO_BABEL
+	bool "Babel"
+	default y
+	help
+	  Babel routing protocol. Distance-vector protocol suitable for
+	  mesh and wireless networks.
+
+config BIRD3_PROTO_RIP
+	bool "RIP"
+	default y
+	help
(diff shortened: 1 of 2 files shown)


tohojo:
Cool idea! Got any numbers on how much space this actually saves for the different variants compared to the full package? :)

PolynomialDivision:
The idea originates from @pktpls from freifunk berlin repository:
https://github.com/freifunk-berlin/falter-packages/commit/66196092e8a7cb192428c71681609f2d05d2a358

There are also size comparisons.

BKPepe:
I don't think it's a good practice for the upstream repository (OpenWrt in this case) to simply rely on a link to a downstream commit for more context, such as package size differences. All relevant information should be included directly in the commit message you are trying to merge here. Furthermore, the linked downstream commit doesn't even seem to contain the size breakdown for the individual packages anyway as you did. 

While splitting the packages sounds like a good idea in theory, it seems we would be the only GNU/Linux distribution doing it this way. Because of that, I wonder if saving a few kilobytes is actually worth it. It will likely cause confusion for users, so I am strongly in favor of making this configurable via Config.in.

PolynomialDivision:
OpenWrt is not a general purpose GNU/Linux distro. It is designed specifically for embedded routers with as little as 8 to 32 MB of flash, where saving around 200 KB in a binary can still matter, even if the final squashfs image reduction is much smaller. The BUILD_VARIANT mechanism exists in OpenWrt specifically for this kind of use case and is already widely used across both the core packages and the various feeds.

I will measure the size differences and include them in the commit message.

PolynomialDivision:
@tohojo is this fine for you to merge?

BKPepe:
Nah, my comment is still not adressed.

Dne pá 10. 7. 2026 8:05 uživatel Nick Hainke ***@***.***>
napsal:

tohojo:
> I will measure the size differences and include them in the commit message.

This bit is still missing? Hard to evaluate without that :)

Also, as for this, you mentioned above:

> It is designed specifically for embedded routers with as little as 8 to 32 MB of flash, where saving around 200 KB in a binary can still matter, even if the final squashfs image reduction is much smaller.

That seems to be a theoretical concern? Do you have any *specific* examples where the size of the Bird binary was a problem for someone building images for one of these small-flash routers? That would strengthen the case for the complexity add here :)

Also (and related), are the exact variants you propose here coming from specific use cases, or was this some "this seems reasonable" kind of selection?

PR #1145, opened 2025-11-10
oonf-*: remove oonf packages such as init scripts, proxy, radio, olsrd2

BKPepe:
These packages have not been compiling for some time due to the transition to GCC14 (issue: https://github.com/OLSR/OONF/issues/71). This has been reported to the upstream repositories, but it appears that development has stalled.

Additionally, the project's website and documentation have been non-functional since 2024. Reference:
https://github.com/OLSR/OONF/issues/54

Another issue is the transition to CMake 4.x, which causes the packages to fail with the following error:
```
CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.
```

Superseeds: https://github.com/openwrt/routing/pull/1144

PR #1176, opened 2026-06-04
oonf-olsrd2: reinstate and update to OONF master

parasew:
## Summary

Reinstates `oonf-olsrd2` (and its `oonf-init-scripts` dependency), which were removed in #1145. The two issues that motivated the removal are both fixed in current OONF master, and I've verified a clean build.

## Why this is safe now

The removal in #1145 cited three problems. Status of each:

- **GCC 14 `-Wint-conversion`** (OLSR/OONF#71): fixed upstream. The `struct msghdr` initializers in `src/base/os_linux/os_system_linux.c` now use designated initializers, which skip musl's padding fields. The fix existed on the `develop` branch and only reached `master` in the 2025-12 `develop`→`master` reconciliation, which is why the old 2022 pin never had it.
- **CMake 4.x minimum version**: resolved. The root `CMakeLists.txt` now sets `cmake_minimum_required(VERSION 3.10 FATAL_ERROR)`, above the 3.5 floor CMake 4.x enforces.
- **Stalled upstream / dead docs** (OLSR/OONF#54): upstream is active again; the README now directs users back to the `master` branch (notice dated 2025-12-04).

## What this PR changes

- Bumps the source pin from the stale `2022-08-25` commit to master `b2164126e12340f19ea33070e1e11eb469a051e5` (2025-12-04), with a refreshed `PKG_MIRROR_HASH`.
- Bumps `PKG_RELEASE`.
- Drops `100-enable-lan-import-plugin.patch`. Upstream has absorbed it: `add_subdirectory(lan_import)` is already present in `src/olsrv2/CMakeLists.txt`, so the patch no longer applies and is redundant.

No source code changes and no plugin-list changes were needed.

## Verification

Clean build and `PKG_MIRROR_HASH` verification with the OpenWrt `x86-64`
`gcc-14.3.0_musl` SDK (snapshots). x86-64 is 64-bit musl, the same ABI condition that triggered the original `-Wint-conversion` failure, so the build exercises the relevant code path. No `-Wint-conversion` and no CMake minimum-version error.

## Scope

This PR covers `oonf-olsrd2` only. The DLEP packages (`oonf-dlep-radio`,
`oonf-dlep-proxy`) were the actual subjects of OLSR/OONF#71 and remain removed; they can follow in a separate PR once each is build-tested individually.

@XDjackieXD noted intent to re-add the packages in #1145.

### DIFF
--- a/oonf-init-scripts/Makefile
+++ b/oonf-init-scripts/Makefile
@@ -1,0 +1,37 @@
+# SPDX-License-Identifier: GPL-2.0-only
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=oonf-init-scripts
+PKG_VERSION:=0.9.1
+PKG_RELEASE:=1
+
+PKG_MAINTAINER:=Matthias Tarasiewicz <mt@riat.at>
+PKG_LICENSE:=GPL-2.0-only
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/oonf-init-scripts
+  SECTION:=net
+  CATEGORY:=Network
+  SUBMENU:=OLSR.org network framework
+  TITLE:=Common OONF startup scripts
+  PKGARCH:=all
+  URL:=https://www.olsr.org/
+  DEPENDS:=+jshn +ubus +uci
+endef
+
+define Package/oonf-init-scripts/description
+  Shared shell helpers sourced by the OONF init scripts.
+endef
+
+define Build/Compile
+endef
+
+define Package/oonf-init-scripts/install
+	$(INSTALL_DIR) $(1)/lib/functions
+	$(INSTALL_DATA) ./files/oonf_init.sh    $(1)/lib/functions/oonf_init.sh
+	$(INSTALL_DATA) ./files/oonf_hotplug.sh $(1)/lib/functions/oonf_hotplug.sh
+endef
+
+$(eval $(call BuildPackage,oonf-init-scripts))
--- a/oonf-init-scripts/files/oonf_hotplug.sh
+++ b/oonf-init-scripts/files/oonf_hotplug.sh
@@ -1,0 +1,10 @@
+#!/bin/sh
+
+case "${ACTION}" in
+    ifup)
+		[ -x "/etc/init.d/${DAEMON}" ] && /etc/init.d/${DAEMON} enabled && {
+			logger -t "${DAEMON}[hotplug]" -p daemon.info 'reloading configuration'
+			/etc/init.d/${DAEMON} reload
+		}
+	;;
+esac
--- a/oonf-init-scripts/files/oonf_init.sh
+++ b/oonf-init-scripts/files/oonf_init.sh
@@ -1,0 +1,136 @@
+#!/bin/sh
+
+. /usr/share/libubox/jshn.sh
+
+oonf_log()
+{
+  logger -s -t "${DAEMON}" -p daemon.info "${1}"
+}
+
+oonf_get_layer3_device()
+{
+  local interface="${1}"  # e.g. 'mywifi'
+  local status dev proto
+  local query="{ \"interface\" : \"${interface}\" }"
+
+  status="$( ubus -S call network.interface status "${query}" )" && {
+    json_load "${status}"
+    json_get_var 'dev' l3_device
+    json_get_var 'proto' proto
+    case "${proto}" in
+      pppoe)
+        # TODO: otherwise it segfaults
+        oonf_log "refusing to add '$interface', because of proto '${proto}'"
+      ;;
+      *)
+        echo "${dev}" # e.g. 'wlan0-1'
+      ;;
+    esac
+  }
+}
+
+oonf_add_devices_to_configuration()
+{
+  local i=0
+  local device_name= section= interface= single_interface=
+  local ignore_list=
+
+  # make a copy of configuration and
+  # add a 'name' (physical name) for all
+  # 'interface-names' (e.g. mywifi)
+
+  # /var is in ramdisc/tmpfs
+  uci export ${DAEMON} >"/var/run/${DAEMON}_dev"
+
+  while section="$( uci -q -c /etc/config get "${DAEMON}.@[${i}]" )"; do {
+    interface="$( uci -q -c /etc/config get "${DAEMON}.@[${i}].ifname" )" || {
+      i=$(( i + 1 ))
+      continue
+    }
+
+    case "$( uci -q get "${DAEMON}.@[${i}].ignore" )" in
+      1|on|true|enabled|yes)
+        oonf_log "ignoring section '${section}'"
+        ignore_list="${i} ${ignore_list}"
+        i=$(( i + 1 ))
+        continue
+      ;;
+    esac
+
+    for single_interface in ${interface}; do {
+      device_name="$( oonf_get_layer3_device "${single_interface}" )"
+
+      if [ ! -z "${device_name}" ]
+      then
+        # add option 'name' for 'ifname' (e.g. 'mywifi')
+        uci -q -c /var/run add_list "${DAEMON}_dev.@[${i}].name=${device_name}"
+      fi
+    } done
+    i=$(( $i + 1 ))
+  } done
+
+  for j in ${ignore_list}; do
+    uci -q -c /var/run delete "${DAEMON}_dev.@[${j}]"
+  done
+
+  uci -q -c /var/run commit "${DAEMON}_dev"
+
+  oonf_log "wrote '/var/run/${DAEMON}_dev'"
+}
+
+
+oonf_reread_config()
+{
+  local pid
+  local pidfile="/var/run/${DAEMON}.pid"
+
+  if   [ -e "${pidfile}" ]; then
+    read pid <"${pidfile}"
+  elif pidfile="$( uci -q get "${DAEMON}.@global[0].pidfile" )"; then
+    read pid <"${pidfile}"
+  fi
+
+  # if empty, ask kernel
+  pid="${pid:-$( pidof ${DAEMON} )}"
+
+  [ -n "${pid}" ] && kill -SIGHUP ${pid}
+}
+
+start()
+{
+  oonf_add_devices_to_configuration
+
+  # produce coredumps
+  ulimit -c unlimited
+
+  service_start /usr/sbin/${DAEMON} --set global.fork=true --load uci:///var/run/${DAEMON}_dev
+}
+
+stop()
+{
+  service_stop /usr/sbin/${DAEMON}
+}
+
+reload()
+{
+  oonf_add_devices_to_configuration
+  oonf_reread_config
+}
+
+running()
+{
+  # check if we have a pidfile and then check if that pid still exists.
+  # since we don't use -e this has to be explicitly returned. exit would stop the process.
+  local pidfile="/var/run/${DAEMON}.pid"
+  test -e "${pidfile}" && test -e "/proc/$(cat "${pidfile}")" && return 0
+  return 1
+}
+
+status()
+{
+  if running; then
+    echo "running"
+  else
+    echo "stopped"
+  fi
+}
--- a/oonf-olsrd2/Makefile
+++ b/oonf-olsrd2/Makefile
@@ -1,0 +1,95 @@
+# SPDX-License-Identifier: GPL-2.0-only
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=oonf-olsrd2
+PKG_RELEASE:=1
+
+PKG_MAINTAINER:=Matthias Tarasiewicz <mt@riat.at>
+PKG_LICENSE:=BSD-3-Clause
+PKG_LICENSE_FILES:=LICENSE
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL:=https://github.com/OLSR/OONF.git
+PKG_SOURCE_DATE:=2026-08-06
+PKG_SOURCE_VERSION:=477e716110ee210f9ce03a75ecacd55a06fd600f
+PKG_MIRROR_HASH:=06fe29b22d2286e8809f17727f0789152148535d212c0b785828b6678f71e254
+
+include $(INCLUDE_DIR)/package.mk
+include $(INCLUDE_DIR)/cmake.mk
+
+
+# ref https://stackoverflow.com/a/10571900/3990041
+SPACE:=$(subst ,, )
+CMAKE_OPTIONAL_PLUGINS:=$(subst $(SPACE),;,$(strip \
+        $(if $(filter y,$(CONFIG_OONF_NHDP_AUTOLL4)),auto_ll4,) \
+        $(if $(filter y,$(CONFIG_OONF_OLSRV2_LAN_IMPORT)),lan_import,) \
+        $(if $(filter y,$(CONFIG_OONF_OLSRV2_ROUTE_MODIFIER)),route_modifier,) \
+        $(if $(filter y,$(CONFIG_OONF_GENERIC_DLEP_ROUTER)),dlep,) \
+        $(if $(filter y,$(CONFIG_OONF_GENERIC_REMOTECONTROL)),remotecontrol,) \
+        $(if $(filter y,$(CONFIG_OONF_OLSRV2_MPR)),mpr,) \
+        $(if $(filter y,$(CONFIG_OONF_GENERIC_HTTP)),http,) \
+        $(if $(filter y,$(CONFIG_OONF_OLSRV2_LAN)),olsrv2_lan,) \
+        $(if $(filter y,$(CONFIG_OONF_OLSRV2_OLD_LAN)),olsrv2_old_lan,) \
+    ))
+
+BUILD_TYPE:=$(if $(filter y,$(CONFIG_DEBUG)),Debug,Release)
+
+CMAKE_OPTIONS+=-D CMAKE_BUILD_TYPE:String=$(BUILD_TYPE) \
+               -D OONF_NO_WERROR:Bool=true \
+               -D OONF_LOGGING_LEVEL:String=debug \
+               -D OONF_NO_TESTING:Bool=true \
+               -D UCI:Bool=true \
+               -D OONF_APP_DEFAULT_CFG_HANDLER:String=uci \
+               -D OONF_STATIC_PLUGINS:String="class;callback;clock;duplicate_set;layer2;packet_socket;rfc5444;socket;stream_socket;telnet;timer;viewer;os_clock;os_fd;os_interface;os_routing;os_system;nhdp;olsrv2;ff_dat_metric;neighbor_probing;nl80211_listener;link_config;layer2info;systeminfo;cfg_uciloader;cfg_compact;nhdpinfo;olsrv2info;netjsoninfo;${CMAKE_OPTIONAL_PLUGINS}" \
+               -D OONF_LIB_GIT:String=$(PKG_SOURCE_VERSION) \
+               -D VERSION_SUB_TAG:String=$(PKG_SOURCE_DATE) \
+               -D INSTALL_LIB_DIR:Path=lib/oonf \
+               -D INSTALL_INCLUDE_DIR:Path=include/oonf \
+               -D INSTALL_CMAKE_DIR:Path=lib/oonf \
+               -D CMAKE_PREFIX_PATH=$(STAGING_DIR)/usr \
+               -D CMAKE_GENERATOR=Ninja
+
+define Package/oonf-olsrd2
+  SECTION:=net
+  CATEGORY:=Network
+  SUBMENU:=OLSR.org network framework
+  TITLE:=OLSRv2 routing agent (olsrd2)
+  URL:=https://www.olsr.org/
+  DEPENDS:=+librt +libnl-tiny +libnl-genl +libuci +oonf-init-scripts
+  MENU:=1
+endef
+
+define Package/oonf-olsrd2/config
+	source "$(SOURCE)/Config.in"
+endef
+
+Build/Compile=$(call Build/Compile/Default,olsrd2_static)
+
+Build/Install=
+
+define Build/Install
+	$(INSTALL_BIN) -D $(PKG_BUILD_DIR)/$(MAKE_PATH)/olsrd2_static $(PKG_INSTALL_DIR)/usr/sbin/olsrd2;
+endef
+
+TARGET_CFLAGS += -I$(STAGING_DIR)/usr/include -I${STAGING_DIR}/usr/include/libnl-tiny -I${STAGING_DIR}/usr/include/libnl3
+
+define Package/oonf-olsrd2/install
+	$(INSTALL_DIR) $(1)/usr/sbin $(1)/etc/init.d $(1)/etc/hotplug.d/iface $(1)/etc/config
+	$(INSTALL_BIN)  $(PKG_BUILD_DIR)/olsrd2_static $(1)/usr/sbin/olsrd2
+	$(INSTALL_BIN)  ./files/olsrd2.init     $(1)/etc/init.d/olsrd2
+	$(INSTALL_DATA) ./files/olsrd2.hotplug  $(1)/etc/hotplug.d/iface/50-olsrd2
+	$(INSTALL_CONF) ./files/olsrd2.uci      $(1)/etc/config/olsrd2
+endef
+
+define Package/oonf-olsrd2/postinst
+#!/bin/sh
+[ -z "$$IPKG_INSTROOT" ] && /etc/init.d/olsrd2 enable
+exit 0
+endef
+
+define Package/oonf-olsrd2/conffiles
+/etc/config/olsrd2
+endef
+
+$(eval $(call BuildPackage,oonf-olsrd2))
--- a/oonf-olsrd2/test-version.sh
+++ b/oonf-olsrd2/test-version.sh
@@ -1,0 +1,18 @@
+#!/bin/sh
+
+# shellcheck shell=busybox
+
+case "$PKG_NAME" in
+oonf-olsrd2)
+	# No upstream release tag exists for the tracked OONF master commit, so
+	# PKG_VERSION is OpenWrt's standard <date>~<short-commit> snapshot string.
+	# The build injects the commit via -D OONF_LIB_GIT, and olsrd2 reports it
+	# in its -v output ("Git commit: <hash>"), so verify the binary reports the
+	# pinned commit rather than skipping the check.
+	olsrd2 -v 2>&1 | grep -F "${PKG_VERSION##*~}"
+	;;
+*)
+	echo "Untested package: $PKG_NAME" >&2
+	exit 1
+	;;
+esac
--- a/oonf-olsrd2/Config.in
+++ b/oonf-olsrd2/Config.in
@@ -1,0 +1,59 @@
+	# OONF Olsrd2 configuration
+	menu "Optional Plugins"
+		depends on PACKAGE_oonf-olsrd2
+
+	config OONF_NHDP_AUTOLL4
+		bool "Auto_LL4 plugin enabled"
+		help
+			The auto_ll4 plugin automatically generates linklocal IPv4 addresses on interfaces that do not contain IPv4 addresses.
+		default n
+
+	config OONF_OLSRV2_LAN_IMPORT
+		bool "Lan_import plugin enabled"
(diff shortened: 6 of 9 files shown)


BKPepe:
Given that the package was removed nearly a year ago and we want to add it back, it basically needs a complete rebirth at this point. We should approach it and review it from scratch to ensure it’s clean and makes sense. If we don’t do it now, it’ll never happen.

I was hoping to convince Copilot to take a look at it, but apparently, it’s a no. :) We’ll see, maybe it’ll get moving. As a next step, I suggest we set up a quick review of the original requirements so we don't bring back any old technical debt.

oonf-init-scripts/Makefile
```diff
@@ -0,0 +1,36 @@
+include $(TOPDIR)/rules.mk
+include $(INCLUDE_DIR)/kernel.mk
+
+PKG_NAME:=oonf-init-scripts
+PKG_VERSION:=0.9.1-r3
+PKG_RELEASE:=1
+PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)
```
BKPepe: Is there a specific reason we need to modify PKG_BUILD_DIR and override the default value?

oonf-init-scripts/Makefile
```diff
@@ -0,0 +1,36 @@
...
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/oonf-init-scripts
+	SECTION:=net
+	CATEGORY:=Network
+	MAINTAINER:=Henning Rogge <hrogge@gmail.com>
```
BKPepe: Does anyone know if this maintainer is still active?

oonf-init-scripts/Makefile
```diff
@@ -0,0 +1,36 @@
...
+	SUBMENU:=OLSR.org network framework
+	URL:=http://www.olsr.org/
```
BKPepe: Would it be possible to switch to HTTPS here?

oonf-init-scripts/Makefile
```diff
@@ -0,0 +1,36 @@
...
+	TITLE:= Common OONF startup scripts
+	VERSION:=$(PKG_VERSION)
```
BKPepe: Why are we explicitly defining the version when it's already the same? This seems redundant.

oonf-init-scripts/Makefile
```diff
@@ -0,0 +1,36 @@
...
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/oonf-init-scripts
+	SECTION:=net
+	CATEGORY:=Network
+	MAINTAINER:=Henning Rogge <hrogge@gmail.com>
+	SUBMENU:=OLSR.org network framework
+	URL:=http://www.olsr.org/
+	TITLE:= Common OONF startup scripts
```
BKPepe:
```suggestion
	TITLE:=Common OONF startup scripts
```

oonf-init-scripts/Makefile
```diff
@@ -0,0 +1,36 @@
...
+	VERSION:=$(PKG_VERSION)
+endef
```
BKPepe: No depends?

oonf-init-scripts/Makefile
```diff
@@ -0,0 +1,36 @@
...
+
+define Build/Configure
+endef
+
+define Build/Compile
+endef
+
+define Package/oonf-init-scripts/install
+	$(INSTALL_BIN)  -D ./files/oonf_init.sh            $(1)/lib/functions/oonf_init.sh
+	$(INSTALL_BIN)  -D ./files/oonf_hotplug.sh         $(1)/lib/functions/oonf_hotplug.sh
```
BKPepe: Could you clarify the purpose of the -d parameter here?

oonf-olsrd2/Makefile
```diff
@@ -0,0 +1,88 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=oonf-olsrd2
+PKG_RELEASE:=3
```
BKPepe:
```suggestion
PKG_RELEASE:=1
```

oonf-olsrd2/Makefile
```diff
@@ -0,0 +1,88 @@
...
+               -D CMAKE_PREFIX_PATH=$(STAGING_DIR)/usr \
+               -D CMAKE_GENERATOR=Ninja
+
+define Package/oonf-git/template
+	SECTION:=net
+	CATEGORY:=Network
+	MAINTAINER:=Henning Rogge <hrogge@gmail.com>
+	SUBMENU:=OLSR.org network framework
+	URL:=http://www.olsr.org/
+	MENU:=1
```
BKPepe: Why menu? It is used? I think it is going to be empty.

oonf-olsrd2/Makefile
```diff
@@ -0,0 +1,88 @@
...
+               -D INSTALL_CMAKE_DIR:Path=lib/oonf \
+               -D CMAKE_PREFIX_PATH=$(STAGING_DIR)/usr \
+               -D CMAKE_GENERATOR=Ninja
+
+define Package/oonf-git/template
+	SECTION:=net
+	CATEGORY:=Network
+	MAINTAINER:=Henning Rogge <hrogge@gmail.com>
+	SUBMENU:=OLSR.org network framework
+	URL:=http://www.olsr.org/
```
BKPepe: Can we use HTTPS here?

oonf-olsrd2/Makefile
```diff
@@ -0,0 +1,88 @@
...
+               -D INSTALL_LIB_DIR:Path=lib/oonf \
+               -D INSTALL_INCLUDE_DIR:Path=include/oonf \
+               -D INSTALL_CMAKE_DIR:Path=lib/oonf \
+               -D CMAKE_PREFIX_PATH=$(STAGING_DIR)/usr \
+               -D CMAKE_GENERATOR=Ninja
+
+define Package/oonf-git/template
+	SECTION:=net
+	CATEGORY:=Network
+	MAINTAINER:=Henning Rogge <hrogge@gmail.com>
```
BKPepe: Same question about the maintainer here.

oonf-olsrd2/Makefile
```diff
@@ -0,0 +1,88 @@
...
+	SUBMENU:=OLSR.org network framework
+	URL:=http://www.olsr.org/
+	MENU:=1
+endef
+
+define Package/oonf-olsrd2
+	$(call Package/oonf-git/template)
+	TITLE:= Build Olsrd V2 Routing Agent
+	DEPENDS:=+librt +libnl-tiny +libuci +oonf-init-scripts
+	VERSION:=$(PKG_VERSION)
```
BKPepe: Here is the question about the version as above.

oonf-olsrd2/Makefile
```diff
@@ -0,0 +1,88 @@
...
+	$(INSTALL_BIN) -D $(PKG_BUILD_DIR)/$(MAKE_PATH)/olsrd2_static $(PKG_INSTALL_DIR)/usr/sbin/olsrd2;
+endef
+
+TARGET_CFLAGS += -I$(STAGING_DIR)/usr/include -I${STAGING_DIR}/usr/include/libnl-tiny
+
+define Package/oonf-olsrd2/install
+	$(INSTALL_BIN)  -D $(PKG_BUILD_DIR)/olsrd2_static $(1)/usr/sbin/olsrd2
+	$(INSTALL_BIN)  -D ./files/olsrd2.init            $(1)/etc/init.d/olsrd2
+	$(INSTALL_BIN)  -D ./files/olsrd2.hotplug         $(1)/etc/hotplug.d/iface/50-olsrd2
+	$(INSTALL_DATA) -D ./files/olsrd2.uci             $(1)/etc/config/olsrd2
```
BKPepe: Could you clarify the purpose of the -d parameter here?

BKPepe: This should be INSTALL_CONF, isnt it?

oonf-olsrd2/Makefile
```diff
@@ -0,0 +1,88 @@
...
+define Package/oonf-olsrd2/config
+	source "$(SOURCE)/Config.in"
+endef
+
+Build/Compile=$(call Build/Compile/Default,olsrd2_static)
+Build/Install=
+
+define Build/Install
+	$(INSTALL_BIN) -D $(PKG_BUILD_DIR)/$(MAKE_PATH)/olsrd2_static $(PKG_INSTALL_DIR)/usr/sbin/olsrd2;
+endef
```
BKPepe, line 72: Why do we need this?

oonf-olsrd2/test-version.sh
```diff
@@ -0,0 +1,18 @@
...
+
+# shellcheck shell=busybox
+
+case "$PKG_NAME" in
+oonf-olsrd2)
+	# PKG_VERSION is a synthetic <date>~<short-commit> string built from
+	# PKG_SOURCE_DATE/PKG_SOURCE_VERSION. olsrd2 reports the OONF library
+	# version via -v/--version (see src/libcore/oonf_main.c), which does not
+	# contain that synthetic string, so the generic version check cannot
+	# match. Skip it explicitly.
```
BKPepe: Im not entirely comfortable with this comment. Why does our versioning diverge from upstream, and how should we verify it? The logic here basically means: we have generic version testing, upstream provides one version, downstream has a different one, and because they don't match, we just completely bypass the generic CI/CD checks. That doesn't seem like the right approach, does it?

BKPepe (changes requested):

parasew:
Agreed, let's treat it as a fresh package rather than a revert. I'll go through both Makefiles against current conventions and push a cleaned version. For the requirements review: the default plugin set is driven by Config.in (lan_import, route_modifier, mpr, etc. as compile-time options folded into the single static binary), oonf-init-scripts ships two sourced shell helpers to /lib/functions/, and oonf-olsrd2 carries the procd init + hotplug + a default UCI config. Happy to align on what the default-enabled plugin set should be before I resubmit, so we don't bring back anything stale.

parasew:
Pushed a cleaned-up version treating this as a fresh package per your suggestion, rather than a revert. Summary of changes: dropped the PKG_BUILD_DIR override, kernel.mk include, redundant VERSION lines, and the oonf-git/template indirection; added SPDX headers, PKG_MAINTAINER, PKG_LICENSE; switched URLs to https; fixed install macros (INSTALL_DIR + INSTALL_CONF for the config file); reset PKG_RELEASE to 1; added dependencies. Replaced the test-version.sh skip with a real check. Fixed several pre-existing bugs in oonf_init.sh that Copilot caught. I built and validated everything in QEMU on the x86-64 snapshot target: the package installs, jshn resolves as a dependency, olsrd2 reports the pinned commit, and the ignore-section logic works correctly. Details in the per-line replies.

oonf-init-scripts/Makefile
parasew: No reason, leftover boilerplate. Removed the override and the Build/Prepare mkdir that referenced it; the default build dir is correct for a files-only package.

oonf-init-scripts/Makefile
parasew: I can't confirm whether Henning is still active on the packaging side. To avoid these going unmaintained again, I'm willing to take over maintenance of oonf-olsrd2 and oonf-init-scripts and have set PKG_MAINTAINER accordingly. I've also emailed Henning directly as a courtesy in case he'd like to stay on. I have several OpenWrt devices and will run olsrd2 on real hardware, not just the cross-compile and QEMU validation I did for this PR.

oonf-init-scripts/Makefile
parasew: Done, switched to https.

oonf-init-scripts/Makefile
parasew: Removed, it duplicated PKG_VERSION.

oonf-init-scripts/Makefile
parasew: Fixed (also reworded the olsrd2 TITLE to drop the odd "Build" verb).

oonf-init-scripts/Makefile
parasew: Added DEPENDS. The helpers source jshn.sh and call ubus/uci, so DEPENDS:=+jshn +ubus +uci. Confirmed on the snapshot target that jshn owns /usr/share/libubox/jshn.sh (apk info --who-owns reports jshn-2026.05.23).

oonf-init-scripts/Makefile
parasew: -D (uppercase) is install's create-leading-directories flag. I've switched to the idiomatic form: an explicit `$(INSTALL_DIR)` followed by plain `$(INSTALL_*)` without -D.

oonf-olsrd2/Makefile
parasew: Done, reset to 1 for the fresh reinstate.

oonf-olsrd2/Makefile
parasew: MENU is actually used here: oonf-olsrd2 has a Config.in plugin-selection submenu wired via Package/oonf-olsrd2/config, and MENU:=1 is what exposes it in menuconfig, so it's not empty. I did remove the oonf-git/template indirection though, since it had a single consumer, and inlined it into the package definition (keeping MENU:=1).

oonf-olsrd2/Makefile
parasew: Yes, switched to INSTALL_CONF for /etc/config/olsrd2 (it's already listed in conffiles).

oonf-olsrd2/Makefile
parasew: I tried removing this (along with CMAKE_INSTALL), but it breaks the build: without it, os_vif_linux.c (pulled in by the lan_import plugin) fails to compile with a "redefinition of struct ethhdr" error under the musl toolchain. So the cmake install handling is load-bearing here, not dead code, and I've kept it. I haven't traced the exact mechanism by which CMAKE_INSTALL affects that compile, but it's empirically required for a clean build across the CI architectures.

oonf-olsrd2/test-version.sh
parasew: You're right that bypassing the check was wrong. The version diverges only because there's no upstream release tag to track, so this follows OpenWrt's standard <date>~<commit> snapshot convention for the pinned master commit. But the build injects that commit via -D OONF_LIB_GIT, and olsrd2 reports it in its -v output. So I replaced the skip with a real check that verifies the binary reports the pinned commit. Validated in QEMU: olsrd2 -v prints "Git commit: b2164126e12340...", which the check matches.

oonf-olsrd2/Makefile
parasew: Removed, it duplicated PKG_VERSION

oonf-olsrd2/Makefile
parasew: Done, switched to https://www.olsr.org/.

oonf-olsrd2/Makefile
parasew: Same as above: I've set PKG_MAINTAINER to myself and emailed Henning as a courtesy. Also removed the oonf-git/template indirection and inlined it into the package definition, so there's now a single place for this.

oonf-olsrd2/Makefile
parasew: -D (uppercase) is the coreutils install create-leading-directories flag. I've switched to the idiomatic form: an explicit `$(INSTALL_DIR)` for the target dirs followed by plain `$(INSTALL_BIN)/$(INSTALL_DATA)/$(INSTALL_CONF)` without -D.

oonf-olsrd2/Makefile
```diff
@@ -0,0 +1,88 @@
...
+               -D UCI:Bool=true \
+               -D OONF_APP_DEFAULT_CFG_HANDLER:String=uci \
+               -D OONF_STATIC_PLUGINS:String="class;callback;clock;duplicate_set;layer2;packet_socket;rfc5444;socket;stream_socket;telnet;timer;viewer;os_clock;os_fd;os_interface;os_routing;os_system;nhdp;olsrv2;ff_dat_metric;neighbor_probing;nl80211_listener;link_config;layer2info;systeminfo;cfg_uciloader;cfg_compact;nhdpinfo;olsrv2info;netjsoninfo;${CMAKE_OPTIONAL_PLUGINS}" \
+               -D OONF_LIB_GIT:String=$(PKG_SOURCE_VERSION) \
+               -D VERSION_SUB_TAG:String=$(PKG_SOURCE_DATE) \
+               -D INSTALL_LIB_DIR:Path=lib/oonf \
+               -D INSTALL_INCLUDE_DIR:Path=include/oonf \
+               -D INSTALL_CMAKE_DIR:Path=lib/oonf \
+               -D CMAKE_PREFIX_PATH=$(STAGING_DIR)/usr \
+               -D CMAKE_GENERATOR=Ninja
```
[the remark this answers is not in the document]
parasew: cmake accepts both -DVAR=value and -D VAR=value; the space form is documented and used elsewhere in the feed. The package builds cleanly across all CI architectures as-is, so this isn't an error. Left the spacing unchanged.

oonf-olsrd2/Makefile
```diff
@@ -0,0 +1,88 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=oonf-olsrd2
+PKG_RELEASE:=3
+
```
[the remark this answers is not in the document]
parasew: Added: SPDX-License-Identifier header (GPL-2.0-only, matching the rest of the feed), PKG_MAINTAINER, and PKG_LICENSE:=BSD-3-Clause (OONF's license per its source headers).

oonf-init-scripts/Makefile
```diff
@@ -0,0 +1,36 @@
+include $(TOPDIR)/rules.mk
+include $(INCLUDE_DIR)/kernel.mk
+
+PKG_NAME:=oonf-init-scripts
+PKG_VERSION:=0.9.1-r3
+PKG_RELEASE:=1
```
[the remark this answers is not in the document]
parasew: Added the SPDX header, PKG_MAINTAINER, and PKG_LICENSE:=GPL-2.0-only here.

oonf-init-scripts/files/oonf_init.sh
```diff
@@ -0,0 +1,137 @@
...
+
+    case "$( uci -q get "${DAEMON}.@[${i}].ignore" )" in
+      1|on|true|enabled|yes)
+        oonf_log "removing/ignore section '$section'"
+        uci -q -c /var/run delete "${DAEMON}_dev.@[${j}]"
+        i=$(( i + 1 ))
+
+        continue
+      ;;
+    esac
```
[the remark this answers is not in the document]
parasew: Good catch, real bug. ${j} was never defined, and as you noted deleting mid-loop shifts the anonymous section indexes. Note Copilot's suggested fix (dropping the delete) would silently disable the ignore feature, so I didn't take it. Instead I collect the ignored indexes during the loop and delete them afterwards in descending order, which removes the section without disturbing the others. Validated in QEMU with a three-section config where the middle section is ignored: that section is removed from /var/run/olsrd2_dev and the surrounding sections stay intact and correctly named.

oonf-init-scripts/files/oonf_init.sh
```diff
@@ -0,0 +1,137 @@
...
+  # 'interface-names' (e.g. mywifi)
+  #
+  # olsrd2.@interface[2]=interface
+  # olsrd2.@interface[2].ifname='wan lan wlanadhoc wlanadhocRADIO1'
+
+  # /var is in ramdisc/tmpfs
+  uci export ${DAEMON} >"/var/run/${DAEMON}_dev"
+
+  while section="$( uci -q -c /etc/config get "${DAEMON}.@[${i}]" )"; do {
+    echo "section: ${section}"
```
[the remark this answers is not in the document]
parasew: Removed the debug echo.

oonf-init-scripts/files/oonf_init.sh
```diff
@@ -0,0 +1,137 @@
...
+        i=$(( i + 1 ))
+
+        continue
+      ;;
+    esac
+
+    for single_interface in ${interface}; do {
+      device_name="$( oonf_get_layer3_device "${single_interface}" )"
+
+      echo "Interface: ${single_interface} = ${device_name}"
```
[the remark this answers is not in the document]
parasew: Removed the debug echo.

oonf-init-scripts/files/oonf_init.sh
```diff
@@ -0,0 +1,137 @@
...
+  oonf_reread_config
+}
+
+running()
+{
+  # check if we have a pidfile and then check if that pid still exists.
+  # since we don't use -e this has to be explicitly returned. exit would stop the process.
+  test -e "/tmp/run/olsrd2.pid" && test -e "/proc/$(cat "/tmp/run/olsrd2.pid")" && return 0
+  return 1
+}
```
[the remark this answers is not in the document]
parasew: Fixed. running() now uses /var/run/${DAEMON}.pid instead of the hardcoded olsrd2 path, consistent with oonf_reread_config elsewhere in the script.

oonf-init-scripts/Makefile
mwarning: As far as I remember he does not maintain the package anymore. But maybe he responds.

oonf-olsrd2/Makefile
BKPepe: Yeah, but MENU previously was in the oonf-git-template, which was wrong even though it was called by oonf-olsrd2 as you corrently pointed out. 

Now, its better. 👍

oonf-init-scripts/Makefile
parasew: I contacted Henning directly as a courtesy and phrased it so that he only needs to respond if he objects or would like to remain listed as maintainer. If he does, I’ll adjust accordingly. Otherwise, I’m happy to take over maintenance of these packages and keep them tested on real OpenWrt hardware.

oonf-init-scripts/files/oonf_hotplug.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+case "${ACTION}" in
+	ifup)
+		. /etc/rc.common /etc/init.d/${DAEMON} enabled && {
+			logger -t "${DAEMON}[hotplug]" -p daemon.info 'reloading configuration'
+			. /etc/rc.common /etc/init.d/${DAEMON} reload
+		}
```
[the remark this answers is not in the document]
parasew: Good catch. Fixed: the handler now calls /etc/init.d/${DAEMON} directly, guarded by [ -x ... ], instead of sourcing rc.common, so the enabled/reload actions actually execute. This matches the standard OpenWrt hotplug idiom. Syntax-checked and the package rebuilds cleanly.

oonf-init-scripts/Makefile
```diff
@@ -0,0 +1,36 @@
...
+include $(INCLUDE_DIR)/package.mk
+
+define Package/oonf-init-scripts
+  SECTION:=net
+  CATEGORY:=Network
+  SUBMENU:=OLSR.org network framework
+  TITLE:=Common OONF startup scripts
+  URL:=https://www.olsr.org/
+  DEPENDS:=+jshn +ubus +uci
+endef
```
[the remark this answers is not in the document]
parasew: Done, added PKGARCH:=all since this package ships only shell helpers and is architecture-independent. The rebuilt package now reports arch:noarch.

oonf-olsrd2/test-version.sh
```diff
@@ -0,0 +1,18 @@
...
+# shellcheck shell=busybox
+
+case "$PKG_NAME" in
+oonf-olsrd2)
+	# No upstream release tag exists for the tracked OONF master commit, so
+	# PKG_VERSION is OpenWrt's standard <date>~<short-commit> snapshot string.
+	# The build injects the commit via -D OONF_LIB_GIT, and olsrd2 reports it
+	# in its -v output ("Git commit: <hash>"), so verify the binary reports the
+	# pinned commit rather than skipping the check.
+	olsrd2 -v 2>&1 | grep -qF "${PKG_VERSION##*~}"
```
BKPepe:
```suggestion
	olsrd2 -v 2>&1 | grep -F "${PKG_VERSION##*~}"
```

We dont need quiet mode here:-)

oonf-olsrd2/Makefile
```diff
@@ -0,0 +1,88 @@
...
+define Build/Install
+	$(INSTALL_BIN) -D $(PKG_BUILD_DIR)/$(MAKE_PATH)/olsrd2_static $(PKG_INSTALL_DIR)/usr/sbin/olsrd2;
+endef
+
+TARGET_CFLAGS += -I$(STAGING_DIR)/usr/include -I${STAGING_DIR}/usr/include/libnl-tiny
+
+define Package/oonf-olsrd2/install
+	$(INSTALL_DIR) $(1)/usr/sbin $(1)/etc/init.d $(1)/etc/hotplug.d/iface $(1)/etc/config
+	$(INSTALL_BIN)  $(PKG_BUILD_DIR)/olsrd2_static $(1)/usr/sbin/olsrd2
+	$(INSTALL_BIN)  ./files/olsrd2.init     $(1)/etc/init.d/olsrd2
```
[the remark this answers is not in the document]
parasew: MAKE_PATH is empty for this package so both paths resolve identically and the build works (verified in QEMU). Happy to align them for consistency if you'd prefer.

oonf-init-scripts/files/oonf_init.sh
```diff
@@ -0,0 +1,136 @@
+#!/bin/sh
+
+. /usr/share/libubox/jshn.sh
+
+oonf_log()
+{
+  logger -s -t ${DAEMON} -p daemon.info "${1}"
+}
```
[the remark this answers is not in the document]
parasew: Fixed, quoted.

parasew:
Rebased onto current master as a single clean commit (the merge commit and the Copilot autofix commit are gone — my mistake, I'd used the "Update branch" button).

Addressed from the formality bot:

Makefile := spacing on SPACE, CMAKE_OPTIONAL_PLUGINS and BUILD_TYPE is fixed.
Quoted ${DAEMON} in the logger call (Copilot) — fixed.
PKG_LICENSE_FILES: upstream OONF ships no LICENSE or COPYING file — the BSD-3-Clause terms exist only as per-file headers in the source, so there is nothing for this parameter to reference. I've opened an issue upstream asking them to add one (OLSR/OONF#76) and will add the parameter once it exists. If you'd prefer, I can point it at a representative source header in the meantime, but that seemed worse than omitting it.

On the remaining Copilot comment about MAKE_PATH: it's empty for this package, so $(PKG_BUILD_DIR)/$(MAKE_PATH)/olsrd2_static and $(PKG_BUILD_DIR)/olsrd2_static resolve identically and the build/install works (verified in QEMU). Happy to align them for consistency if you'd rather.

parasew:
On PKG_LICENSE_FILES: upstream confirmed BSD-3-Clause and asked me to submit a LICENSE file (OLSR/OONF#76), which I've now done (OLSR/OONF#77). Once it's merged I can add PKG_LICENSE_FILES:=LICENSE ; this will also require to bumping the source pin to a commit containing the file, plus a rebuild/re-verify.

parasew:
While testing on real hardware (GL.iNet Flint 2, aarch64/filogic), found that nl80211_listener needs libnl-genl-3.0, which libnl-tiny doesn't provide. The existing core netlink code (os_system_linux.c) still uses libnl-tiny at runtime, so libnl-genl is needed in addition to it, not instead of it. Added +libnl-genl to DEPENDS alongside the existing +libnl-tiny, and the libnl3 include path alongside the libnl-tiny one.

Verified building, packaging, and installing cleanly on both x86-64 (gcc-14.3.0_musl SDK) and aarch64_cortex-a53 (mediatek/filogic, OpenWrt 25.12.5), with correct dependency resolution on both. Re-ran the QEMU checks on x86-64: jshn ownership, olsrd2 -v reporting the pinned commit with nl80211_listener now genuinely loaded, and the ignore-section config logic all pass.

parasew:
Bumped the source pin to 477e716 (2026-08-06), which includes the upstream LICENSE file added in OLSR/OONF#77. Added PKG_LICENSE:=BSD-3-Clause and PKG_LICENSE_FILES:=LICENSE accordingly.

parasew:
@BKPepe can you restart the Copilot check please?

parasew:
Just checking in - is there anything left to do (besides re-starting the copilot runs)?

PR #1190, opened 2026-08-06
cjdns: fix version reporting, riscv64 build and CI runtime tests

BKPepe:
This fixes the failing CI runtime tests for cjdns, which is part of the fallback test set (`bird2 cjdns olsrd`) used for every PR that does not change any package, so these failures currently make CI red repo-wide.

- `CJDNS_RELEASE_VERSION` was set to `PKG_SOURCE_VERSION`, which is never defined for tarball builds, so `cjdroute --version` reported `Cjdns version: unknown` and the generic version check failed. Pass `PKG_VERSION` instead.
- The bundled cnacl build system has no premade build plan for riscv64 and aborts (`Error: build with no premade plan, TODO: generate one`), breaking every riscv64 CI job. Disable the package there until upstream gains support.
- Add `test-version.sh`: verify the version via `cjdroute --version` for cjdns and skip the check for cjdns-tests, whose test binary provides no version information.

Maintainer: @wfleurant

### DIFF
--- a/cjdns/Makefile
+++ b/cjdns/Makefile
@@ -37,7 +37,7 @@ define Package/cjdns
   SUBMENU:=Routing and Redirection
   TITLE:=Encrypted near-zero-conf mesh routing protocol
   URL:=https://github.com/cjdelisle/cjdns
-  DEPENDS:=@!arc @IPV6 +kmod-tun +libnl-tiny +libpthread +librt \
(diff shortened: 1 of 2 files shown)


BKPepe:
[allow cherry-pick]

BKPepe:
@wfleurant What do you think about this?

wfleurant:
thanks boss -- strings probably works in that bash script eh? else we fix it in another one.
yeah ship it!

PR #1191, opened 2026-08-06
olsrd: fix plugin SONAME, add test-version.sh, clean up Makefile

BKPepe:
This fixes the failing CI runtime tests for olsrd, which is part of the fallback test set (`bird2 cjdns olsrd`) used for every PR that does not change any package, so these failures currently make CI red repo-wide.

- The plugins are linked with SONAME `lib$(PLUGIN_NAME).so` but installed as `$(PLUGIN_NAME).so.$(PLUGIN_VER)` without any matching symlink, so the library check of the CI runtime tests failed for all 18 olsrd-mod-* packages. Add a patch setting the SONAME to the installed filename; the plugins are dlopen()ed by path, so this has no functional effect. The patch will also be submitted upstream.
- Add `test-version.sh`: the package version is date+hash based and not reported by any binary, so the generic version check could never pass. Verify the olsrd version banner instead and skip the check for plugin/utility packages.
- Makefile cleanup: remove the MAINTAINER override in the shared template which shadowed PKG_MAINTAINER for all packages, add PKG_LICENSE_FILES, use HTTPS homepage URL.

### DIFF
--- a/olsrd/Makefile
+++ b/olsrd/Makefile
@@ -17,6 +17,7 @@ PKG_MIRROR_HASH:=e05d2724e6c76233148ce460b9611bafccff2114904542fa3b5caf75dc1c4e0
 PKG_MAINTAINER:=Nick Hainke <vincent@systemli.org>
 PKG_BUILD_PARALLEL:=0
 PKG_LICENSE:=BSD-3-Clause
+PKG_LICENSE_FILES:=license.txt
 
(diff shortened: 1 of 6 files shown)


BKPepe:
[allow cherry-pick]

BKPepe:
Gentle reminder. :)

Issue #184, opened 2016-05-04
Merging the openwrt-routing/packages with openwrt/packages?


jow-:
Hi @openwrt-routing/owners

for a long time both the openwrt-routing and openwrt packages organizations are working in parallel and in many cases are maintained by the same persons.

For the sake of simplicity I'd like to suggest to merge both organizations and move the packages from here into openwrt/packages.

This issue is just for starting a discussion and maybe reaching a consensus, so please voice your opinion here.

ffainelli:
Seems reasonable to me to merge this within openwrt/packages. AFAIR this project started earlier than when we decided to move the packages maintenance to Github, so no objection from me.

zioproto:
Hello @jow- 
sounds good to me to merge organizations. Less overhead is always better.
openwrt-routing was kind of a small group, where everyone knows each other because of the many OpenWrt and Battlemesh events.
If the group of committers gets wider in the new organization we should find a way to resolve the situation where there is a patch that needs further improvement before it gets accepted.

I have a proposal, can we set up the repositories in a way that is never possible to push directly but it is always necessary a pull request ? In this way we know that each change had at least 1 other person reviewing it.

lynxis:
@zioproto you can protect branches and enable require status checks. it's not the same, but nearby.

lgierth:
No objections from my side

mwarning:
sure, go ahead

bittorf:
i'am against it. it's easier for me to follow in this special field and i dont want to read all the messages from packages.

axn:
I see the point and for me both strategies are fine.

Could it mean less flexibility when interested in bleeding-edge routing
packages from trunk but stable openwrt packages? Now one could easily
choose different branches for each repo. But that's probably not so
relevant.

/axel

elektra42:
Hi –

> I see the point and for me both strategies are fine.
> 
> Could it mean less flexibility when interested in bleeding-edge routing
> packages from trunk but stable openwrt packages? Now one could easily
> choose different branches for each repo. But that's probably not so
> relevant.

+1

Cheers,
Elektra

-- 
Viral meme of radical freedom 
 
The fact that you talk in your head doesn't mean that you think. 
 
The best way to lose control over yourself is trying to control yourself. 
 
Most people experience themselves as a voice in their head, telling them 
 who they are, what they think and what they have to do. 
 
http://en.wikipedia.org/wiki/Meme

aparcar:
[...]

PR #1192, opened 2026-08-06
alfred: clean up Makefile

BKPepe:
Cleanup towards moving the package to openwrt/packages (#184):

- Install binaries with INSTALL_BIN instead of `cp -fpR`.
- Replace `[ "x$(CONFIG_...)" == "xy" ] && ... ; true` shell constructs (the `==` operator is a bashism) with make-level `$(if ...)` conditionals.
- Drop the explicit dependency on libc, which every package depends on implicitly.
- Indent MAKE_FLAGS consistently with tabs.

No functional change in the resulting package.

Maintainer: @simonwunderlich

### DIFF
--- a/alfred/Makefile
+++ b/alfred/Makefile
@@ -4,7 +4,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=alfred
 PKG_VERSION:=2026.2
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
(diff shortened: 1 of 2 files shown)


ecsv:
@BKPepe I think we had that before. @simonwunderlich is the maintainer. And i can't merge anything here.

ecsv (approved):
Changes seem to be fine

PR #1194, opened 2026-08-06
bmx7: clean up Makefile and add test-version.sh

BKPepe:
- Remove `GIT_REV="$(PKG_REV)"` from MAKE_ARGS: PKG_REV is not defined anywhere and the bmx7 build system does not consume a GIT_REV make variable at all.
- Do not create empty /etc/config and /etc/init.d directories in the bmx7 package; the files installed there belong to bmx7-uci-config.
- Add `test-version.sh` for the CI runtime tests: the package version is derived from PKG_SOURCE_DATE, which the binary does not report, so the generic version check could never pass. Running `bmx7 -v` in CI is not suitable either, since it first generates the node RSA-2048 key, which can take very long on emulated architectures.

Maintainer: @axn

### DIFF
--- a/bmx7/test-version.sh
+++ b/bmx7/test-version.sh
@@ -1,0 +1,23 @@
+#!/bin/sh
+
+# shellcheck shell=busybox
+
+case "$PKG_NAME" in
+bmx7)
+	# The version of bmx7 is derived from the source date, which the
+	# binary does not report. Running `bmx7 -v` is not suitable for
+	# CI either, as it first generates the node RSA key, which can
+	# take a very long time on emulated architectures.
+	exit 0
+	;;
+
+bmx7-*)
+	# Plugins are libraries and do not provide version information
+	exit 0
+	;;
+
+*)
+	echo "Untested package: $PKG_NAME" >&2
+	exit 1
+	;;
+esac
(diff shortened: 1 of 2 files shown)


bmx7/test-version.sh
```diff
@@ -0,0 +1,24 @@
...
+
+# shellcheck shell=busybox
+
+case "$PKG_NAME" in
+bmx7)
+	# The version of bmx7 is derived from the source date, which the
+	# binary does not report. `bmx7 -v` is not an alternative: the
+	# key path option is applied before the version option, so a
+	# missing node key gets generated first - "Creating RSA2048
+	# private key. This can take a while", as bmx7 puts it.
```
BKPepe, line 11: @axn — question rather than a bug report: bmx7 -v prints the node identity (key hash, key type, link keys, descSqn), so it needs myKey and therefore generates the RSA2048 node key on first run before printing anything. That is fine for what the option does, but it means there is no cheap way to ask a bmx7 binary which version it is, which is why this PR skips the version check in CI rather than running the binary.

Would a lightweight option that prints just BMX_BRANCH-BRANCH_VERSION and exits be of interest upstream?

BKPepe:
[allow cherry-pick]

BKPepe:
@axn Gentle ping

PR #1195, opened 2026-08-06
batmand, vis: drop dead build flags, fix revision info, add test-version.sh

BKPepe:
- Both packages passed an `EXTRA_CFLAGS='...'` assignment that the upstream build systems never reference, so none of those flags were ever applied (the debug defines are already set via CPPFLAGS upstream and REVISION_VERSION is unused in both sources). Drop it.
- vis: the REVISION make variable was set to the undefined PKG_REV, so the binary carried no revision information. Use PKG_SOURCE_VERSION like batmand does (PKG_RELEASE bumped).
- Add `test-version.sh` for the CI runtime tests: the package versions are derived from PKG_SOURCE_DATE and the git hash, which the binaries do not report, so the generic version check could never pass. Check the version banners via `batmand -v` / `vis -v` instead.

Maintainer: @elektra42


batmand/patches/100-allocate-include-string-h.patch
```diff
@@ -0,0 +1,27 @@
...
+Subject: [PATCH] allocate: include string.h for memcpy
+
+memcpy() is used in debugRealloc() without including string.h. Since
+GCC 14 treats implicit function declarations as a hard error, the
+build fails:
+
+    allocate.c:349:17: error: implicit declaration of function
+    'memcpy' [-Wimplicit-function-declaration]
+
+Signed-off-by: Josef Schlehofer <pepe.schlehofer@gmail.com>
```
BKPepe, line 13: @elektra42 @ecsv:

This PR carries a small patch adding #include <string.h> to allocate.c, because memcpy() is used there without it and GCC 14 turns implicit declarations into errors, so batmand currently fails to build.

It looks like a fix that belongs upstream rather than in the feed.

BKPepe:
[allow cherry-pick]

PR #1197, opened 2026-08-06
babeld: simplify LDLIBS in MAKE_FLAGS

BKPepe:
Setting LDLIBS to an empty string and appending to it on the same command line is equivalent to assigning the final value directly. No change in the resulting package.


BKPepe:
[allow cherry-pick]

BKPepe:
@PolynomialDivision gentle ping

PR #1199, opened 2026-08-06
batman-adv, batctl, mesh11sd, nodogsplash, opennds: fix defects found while reviewing the move to openwrt/packages

BKPepe:
These are defects that surfaced while preparing the move of these packages to openwrt/packages (#184). They are all pre-existing in this feed, so they are worth fixing here rather than only in the migration PRs — those may take a while to land, and some of this affects users today.

One commit per package:

- **batman-adv** — `/etc/config/batman-adv` is registered as a conffile although nothing installs it; the shipped migration script deletes that file as its last step. The same script uses `continue` to skip a section although there is no loop in the function, which only works because ash walks the dynamic call stack into `config_foreach`.
- **batctl** — `PKG_BUILD_DIR` duplicated the default `package.mk` derives for a package with build variants; the `CONFIG_*` switches went through a `shexport` shell variable although listing them on the make command line is equivalent (~2 KB for the largest variant); plus a user-visible typo and inconsistent indentation.
- **mesh11sd** — `mesh11sd -v` waits for `br-lan` through `wait_for_interface()` before printing anything, for the full `interface_timeout` (10 s) when the interface is not up. Patch also submitted upstream as openNDS/mesh11sd#204.
- **nodogsplash** — the init script prints two error messages with the wrong variable, and emits `gatewayinterface` twice, so the raw UCI value can override the device resolved via `network_get_device()`.
- **opennds** — `URL` uses a casing that only resolves through GitHub's redirect.

Each package that changes content gets a `PKG_RELEASE` bump.

Maintainers: @simonwunderlich (batman-adv, batctl), @bluewavenet (mesh11sd, opennds), @mwarning (nodogsplash)

[allow cherry-pick]

### DIFF
--- a/batctl/Makefile
+++ b/batctl/Makefile
@@ -4,12 +4,11 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=batctl
 PKG_VERSION:=2025.4
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://downloads.open-mesh.org/batman/releases/batman-adv-$(PKG_VERSION)
 PKG_HASH:=566aa4fd74355e8d5dd2bd21c25e98cc068e77fd08e19f80b8e2b5edde7c8e65
-PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)-$(BUILD_VARIANT)/$(PKG_NAME)-$(PKG_VERSION)
 
 PKG_MAINTAINER:=Simon Wunderlich <sw@simonwunderlich.de>
 PKG_LICENSE:=GPL-2.0-only ISC MIT
@@ -29,7 +28,7 @@ define Package/batctl/Default
 endef
 
 define Package/batctl/description
-  batctl is a more intuitive managment utility for B.A.T.M.A.N.-Advanced.
+  batctl is a more intuitive management utility for B.A.T.M.A.N.-Advanced.
   It is an easier method for configuring batman-adv and provides some
   additional tools for debugging as well. This package builds
   version $(PKG_VERSION) of the user space utility.
@@ -72,11 +71,11 @@ $(Package/batctl/description)
 endef
 
 MAKE_VARS += \
-        LIBNL_NAME="libnl-tiny" \
-        LIBNL_GENL_NAME="libnl-tiny"
+	LIBNL_NAME="libnl-tiny" \
+	LIBNL_GENL_NAME="libnl-tiny"
 
 MAKE_FLAGS += \
-        REVISION="$(PKG_VERSION)-openwrt-$(PKG_RELEASE)"
+	REVISION="$(PKG_VERSION)-openwrt-$(PKG_RELEASE)"
 
 config-n := \
 	aggregation \
@@ -214,17 +213,9 @@ config-y := \
 
 endif
 
-define ConfigVars
-$(subst $(space),,$(foreach opt,$(config-$(1)),CONFIG_$(opt)=$(1)
-))
-endef
-
-define batctl_config
-$(call ConfigVars,n)$(call ConfigVars,y)
-endef
-$(eval $(call shexport,batctl_config))
-
-MAKE_FLAGS += $$$$$(call shvar,batctl_config)
+MAKE_FLAGS += \
+	$(patsubst %,CONFIG_%=n,$(config-n)) \
+	$(patsubst %,CONFIG_%=y,$(config-y))
 
 define Package/batctl-tiny/install
 	$(INSTALL_DIR) $(1)/usr/libexec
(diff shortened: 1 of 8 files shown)


BKPepe:
@ecsv @simonwunderlich — heads-up on the batman-adv commit in this PR: the problem is not only on the OpenWrt side.

`compat-include/linux/slab.h` in the batman-adv tarball carries the same guard as `src/compat-hacks.h` does here:

```c
#if LINUX_VERSION_IS_LESS(7, 0, 0) && \
    !(LINUX_VERSION_IS_GEQ(6, 18, 33) && LINUX_VERSION_IS_LESS(6, 19, 0))
```

That range only carves out the 6.18.33 stable backport, but `kzalloc_obj` / `kmalloc_obj` / `kmalloc_objs` have landed in 6.12.y as well — 6.12.100 has them — so the definitions collide with `include/linux/slab.h`. I checked `open-mesh-mirror/batman-adv` `main` and the guard is unchanged there, so an out-of-tree build using batman-adv's own Makefile (which adds `-I$(PWD)/compat-include/`) against such a kernel should hit the same error. The OpenWrt package does not use `compat-include/` — it force-includes its own `compat-hacks.h` instead — which is why the same three macros exist in two places and why this PR fixes the copy here.

CI made it concrete: exactly the four targets on `KERNEL_PATCHVER 6.12` failed and the six on 6.18 passed.

```
compat-hacks.h:52:9: error: "kzalloc_obj" redefined [-Werror]
./include/linux/slab.h:963:9: note: this is the location of the previous definition
```

This PR switches the guards to `#ifndef kzalloc_obj` and friends, so the compat code steps aside whenever the kernel provides the helper, without tracking which point release of which stable series picked the backport up. The same change would apply verbatim to `compat-include/linux/slab.h` — that file already does `#include_next <linux/slab.h>` at the top, so it needs only the guard swap and not the extra include this PR adds here.

Since the GitHub mirror has issues and pull requests disabled, I take it patches go to `b.a.t.m.a.n@lists.open-mesh.org`. Would you like me to send the `#ifndef` form there, or would you rather take it upstream directly?

ecsv:
> @ecsv @simonwunderlich — heads-up on the batman-adv commit in this PR: the problem is not only on the OpenWrt side.
> 
> `compat-include/linux/slab.h` in the batman-adv tarball carries the same guard as `src/compat-hacks.h` does here:
> 
> ```c
> #if LINUX_VERSION_IS_LESS(7, 0, 0) && \
>     !(LINUX_VERSION_IS_GEQ(6, 18, 33) && LINUX_VERSION_IS_LESS(6, 19, 0))
> ```
[...]
> ```
> compat-hacks.h:52:9: error: "kzalloc_obj" redefined [-Werror]
> ./include/linux/slab.h:963:9: note: this is the location of the previous definition
> ```

Upstream batman-adv handles it via:

```
#if LINUX_VERSION_IS_LESS(7, 0, 0) && \
    !(LINUX_VERSION_IS_GEQ(6, 18, 33) && LINUX_VERSION_IS_LESS(6, 19, 0)) && \
    !(LINUX_VERSION_IS_GEQ(6, 12, 97) && LINUX_VERSION_IS_LESS(6, 13, 0))
```

> This PR switches the guards to `#ifndef kzalloc_obj` and friends, so the compat code steps aside whenever the kernel provides the helper, without tracking which point release of which stable series picked the backport up. The same change would apply verbatim to `compat-include/linux/slab.h` — that file already does `#include_next <linux/slab.h>` at the top, so it needs only the guard swap and not the extra include this PR adds here.
> 
> Since the GitHub mirror has issues and pull requests disabled

It is just a read-only mirror. It is not officially support and might be completely out-of-date.

Please use https://git.open-mesh.org/batman-adv.git/ as reference

> I take it patches go to `b.a.t.m.a.n@lists.open-mesh.org`.

Yes, see https://www.open-mesh.org/projects/open-mesh/wiki/Contribute#Submitting-patches 

> Would you like me to send the `#ifndef` form there, or would you rather take it upstream directly?

No, I would like to keep the way upstream way for now. You cannot assume that it will stay a define.

We can talk about the `#ifndef`s when you would keep the check to completely disable it for newer kernels

```
#if LINUX_VERSION_IS_LESS(7, 0, 0)
```

ecsv (approved):
For the batctl + batman-adv: Looks mostly ok. But I would prefer when you would keep the outer check to completely disable it for kernels >= 7.0 (`#if LINUX_VERSION_IS_LESS(7, 0, 0)`). Will make it easier to avoid problems when these are no longer defines or to directly see when these compat defines can finally dropped from this file.

ecsv:
> > I take it patches go to `b.a.t.m.a.n@lists.open-mesh.org`.
> 
> Yes, see https://www.open-mesh.org/projects/open-mesh/wiki/Contribute#Submitting-patches

Since you did all the rest, I will handle this part

EDIT: https://git.open-mesh.org/batman-adv.git/commit/?id=96a26bababe5e9c9b9f9226e7cdbe60f49f57b71

ecsv:
@BKPepe can you please rebase it to resolve the conflicts? In theory, then this PR would only be about:

* mesh11sd
* nodogsplash
* opennds

